PR #239, opened 2014-09-03
Update to 2.10

jcoveney:
This begins the process... there is still a little more work that needs to be done. I need CI to run to see how the tests that I don't have set up to run locally run, and also, there's an open question around twitter's kafka fork, and whether that should be exposed in the OSS storehaus.


ianoc:
Looks like this must have already be brought in via another route. Killing

PR #242, opened 2014-11-06
Feature/fanout store

MansurAshraf:
A store that fans out Key K to a Set of Keys, queries the appropriate underlying stores and sum the values using the Monoid V

### DIFF
--- a/storehaus-core/src/main/scala/com/twitter/storehaus/FanoutStore.scala
+++ b/storehaus-core/src/main/scala/com/twitter/storehaus/FanoutStore.scala
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2014 Twitter inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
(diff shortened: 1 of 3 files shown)


storehaus-core/src/main/scala/com/twitter/storehaus/FanoutStore.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.twitter.algebird.{Semigroup, Monoid}
+import com.twitter.util.Future
+
+/**
+ * A store that fans out Key K to a Set of Keys, queries the appropriate underlying stores and sum
+ * the values using the Monoid V
+ *
+ * @author Mansur Ashraf
+ */
+class FanoutStore[-K, +V: Semigroup, S <: ReadableStore[K, V]](fanout: K => Set[K], stores: Set[(K => Boolean, S)])
```
johnynek: This is a cool idea, but question: why not make it:
```scala
class FanoutStore[-K, K1, +V: Semigroup, S <: ReadableStore[K1, V]](fanout: K => Iterable[K1], stores: K1 => Iterable[S])
```
the `Set[(LK => Boolean, S)]` seems confusing to me. It seems you just want a lookup function, right?

storehaus-core/src/main/scala/com/twitter/storehaus/FanoutStore.scala
```diff
@@ -0,0 +1,62 @@
...
+ *    limitations under the License.
+ */
+package com.twitter.storehaus
+
+import com.twitter.algebird.{Semigroup, Monoid}
+import com.twitter.util.Future
+
+/**
+ * A store that fans out Key K to a Set of Keys, queries the appropriate underlying stores and sum
+ * the values using the Monoid V
```
johnynek: Semigroup.

johnynek:
This seems useful. Let's get it merged.

Also, it would be great to describe a concrete use case in the comments to the class. I think it will help people imagine how to use it. I think it is for read-time aggregation of a set of stores in a more general way than summingbird's clientstore works.

PR #244, opened 2014-11-18
Fixes #243 honoring mutable cache ttl-ness in getOrElseUpdate and hit

### LINKED ISSUE #243: MutableTTLCache.hit should call get(k) instead of backingCache.get
When implementing a ReadableStore.withCache(someStore, MutableTTLCache) I noticed that entries were not properly expiring and traced the issue back to this line. Calling hit(k) on the MutableTTLCache seems to call directly into the backing cache (in this case a JLinkedHashMap) and does not apply any of the ttl-ness filtering found in MutableTTLCache.get. Hit then calls +=, which adds the entry back to the cache with a new TTL

aschwager:


clocklear:
I'd like to upvote this PR.  I just had this soft-TTL issue bite me a bit because I didn't understand how it works.  I'd like to see an option to make the behavior controllable at cache creation (either hard or soft TTL-ness).

PR #257, opened 2015-02-24
Add correct String/ChannelBuffer injections

franklinhu:
The String and ChannelBuffer injections provided in storehaus-mysql are flipped (there's always a mapping to MySqlValue but not always back). Added injections in the right direction and deprecated the existing ones.

* Add `String2MySqlValueInjection` and `ChannelBuffer2MySqlValueInjection` 
* Deprecate `MySqlStringInjection` and `MySqlCbInjection`

### DIFF
--- a/storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/ValueMapper.scala
+++ b/storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/ValueMapper.scala
@@ -124,9 +124,15 @@ class MySqlValue(val v: Value) {
  * Returns string representation of the finagle-mysql Value wrapped by MySqlValue
  * Both null values and empty values map to empty string.
  */
+@deprecated("Use String2MySqlValueInjection", "0.10.0")
 object MySqlStringInjection extends Injection[MySqlValue, String] {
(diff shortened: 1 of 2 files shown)


johnynek:
Yes, it does look like the original ones are a bit ill-formed. Can you add some tests here? The IsInjection property is pretty easy to use as long as you have the Gen/Arbitrary for the types in question.

franklinhu:
It looks like IsInjection is only accessible in bijection-core?https://github.com/twitter/bijection/blob/23b7978fd8a4948ad294aa44fda35db5747950bd/bijection-core/src/test/scala/com/twitter/bijection/BaseProperties.scala#L66-L68

It also seems like we could remove the storehaus `MySqlValue` wrapper over the Finagle types since bijection-finagle-mysql exists now?

johnynek:
sadly those BaseProperties are not exported. You'll have to do something like:

```scala
forAll { (i: Int) =>
  injection.invert(injection(i)) == Success(i)
}
```
is good enough.

About bijection-finagle-mysql: yes we should remove the repeated code here (once the tests are passing on that and we have a new published version). For now let's make an issue with the correct links.

franklinhu:
Added the tests, and ended up discovering a bug in how ChannelBuffers are converted to MySQL types.

johnynek:
Thanks for this! Looks good.

Issue #51, opened 2013-03-09
storehaus-leveldb


sritchie:
Docs for implementing get, put and multiPut are pretty good:

https://github.com/fusesource/leveldbjni

Check out the LevelDB persistence implementation in ElephantDB for clues:

https://github.com/nathanmarz/elephantdb/blob/develop/elephantdb-leveldb/src/jvm/elephantdb/persistence/LevelDB.java

PR #258, opened 2015-03-13
LevelDB support

BenFradet:
As per #51, this is the first draft for supporting LevelDB inside storehaus, feedback is more than welcome.

I'm planning on doing `multiGet` and `multiPut` in the coming week.

I intend to do the same thing with RocksDB once I'm done here.


storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,86 @@
...
+      case _ => None
+    }
+  }
+
+  /** Get a set of keys from the store.
+    * Important: all keys in the input set are in the resulting map. If the
+    * store fails to return a value for a given key, that should be represented
+    * by a Future.exception.
+    */
+  override def multiGet[K1 <: Array[Byte]](ks: Set[K1])
```
rubanm: This is defined in the parent class so not needed here right? Likewise for `multiPut` and `close(time: Time)`

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  /** get a single key from the store.
+    * Prefer multiGet if you are getting more than one key at a time
+    */
+  override def get(k: Array[Byte]): Future[Option[Array[Byte]]] = futurePool {
+    require(k != null)
+    db.get(k) match {
+      case a: Array[Byte] => Some(a)
+      case _ => None
+    }
```
rubanm: This can be just `Option(db.get(k))`? given absent keys return null

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,86 @@
...
+  /** Replace a set of keys at one time */
+  override def multiPut[K1 <: Array[Byte]](kvs: Map[K1, Option[Array[Byte]]])
+    : Map[K1, Future[Unit]] = super.multiPut(kvs)
+
+  /** Close this store and release any resources.
+    * It is undefined what happens on get/multiGet after close
+    */
+  override def close(time: Time): Future[Unit] = super.close(time)
+
+  override def close(after: Duration): Future[Unit] = {
```
rubanm: Which one does this override? Don't see a version that takes in `Duration`. Should this just take in `Time`?

storehaus-testing/src/main/scala/com/twitter/storehaus/testing/generator/NonEmpty.scala
```diff
@@ -22,12 +22,20 @@ import org.scalacheck.{ Choose, Gen }
...
-    for(cs <- Gen.listOf1(Gen.alphaChar)) yield cs.mkString
+    for (cs <- Gen.listOf1(Gen.alphaChar)) yield cs.mkString
 
   /** Generator for Options of non-empty alpha strings of random length */
   def alphaStrOpt: Gen[Option[String]] =
     alphaStr.flatMap(str => Gen.oneOf(Some(str), None))
 
+  /** Generator for non-empty byte arrays of random length */
+  def byteArray: Gen[Array[Byte]] =
+    for (cs <- Gen.listOf1(Gen.alphaChar)) yield cs.mkString.getBytes
```
rubanm: Should we force an encoding here?

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
rubanm: Ah, please disregard. Just saw your PR comment about implementing this.

rubanm:
@BenFradet Thanks for the PR. Mostly minor comments on this version.

BenFradet:
Thanks a lot for your feedback, will take it into account.

BenFradet:
All unit tests for the module are passing.

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.twitter.util.{FuturePool, Time, Future, Duration}
+import org.iq80.leveldb._
+import org.fusesource.leveldbjni.JniDBFactory._
+
+/**
+ * Store dealing with a LevelDB database.
+ * @author Ben Fradet
+ * @since 10/03/15
+ */
+class LevelDBStore(val dir: File, val options: Options, val numThreads: Int)
```
rubanm: numThreads can be defaulted to Runtime.getRuntime.availableProcessors perhaps. Some of the other store wrappers that use future pool do this.

storehaus-leveldb/src/test/scala/com/twitter/storehaus/leveldb/LevelDBStoreProperties.scala
```diff
@@ -0,0 +1,73 @@
...
+        }
+      }
+    }
+
+  def multiPutAndGetTest(store: Store[Array[Byte], Array[Byte]],
+                         pairs: Gen[List[(Array[Byte], Option[Array[Byte]])]]) =
+    forAll(pairs) { examples: List[(Array[Byte], Option[Array[Byte]])] =>
+      val examplesMap = examples.toMap
+      Await.result(Future.collect(store.multiPut(examplesMap).values.toList))
+      val result = store.multiGet(examplesMap.keySet).map(kv => kv match {
```
rubanm: This can just be .map { case (key, v) => ... } right?

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,83 @@
...
+import java.io.File
+import java.util.concurrent.Executors
+
+import com.twitter.storehaus.Store
+import com.twitter.util.{FuturePool, Time, Future, Duration}
+import org.iq80.leveldb._
+import org.fusesource.leveldbjni.JniDBFactory._
+
+/**
+ * Store dealing with a LevelDB database.
```
rubanm: Could you add some more comments about this store's usage, what the parameters mean etc. Thanks.

storehaus-leveldb/src/test/scala/com/twitter/storehaus/leveldb/LevelDBStoreProperties.scala
```diff
@@ -0,0 +1,73 @@
...
+      val stringifiedExamples = stringifyMap(examplesMap)
+
+      stringifiedResults == stringifiedExamples
+    }
+
+  private def stringifyMap(map: Map[Array[Byte], Option[Array[Byte]]])
+  :Map[String, Option[String]] = {
+    map.map(kv => kv match {
+      case (k, Some(v)) => (new String(k), Some(new String(v)))
+      case (k, None) => (new String(k), None)
```
rubanm: need to also specify the encoding here?

rubanm:
@BenFradet Looks good to me. A few minor comments.

There was an ivy issue with memcache tests (hopefully transient). I've restarted the build.

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,83 @@
...
+ * @author Ben Fradet
+ * @since 10/03/15
+ */
+class LevelDBStore(val dir: File, val options: Options, val numThreads: Int)
+    extends Store[Array[Byte], Array[Byte]] {
```
rubanm: We could make this Store[K, V] and and accept an implicit Codec[K] and Codec[V].

Not a blocker though given we have a combinator that can be used instead. ConvertedStore I think.

johnynek: Actually, I tend to prefer wrapping as close to the underlying store as possible and using combinators (I know @ianoc has some information that can cause problems, but otherwise we have a ton of duplication, which I also think is a problem). What do you think of this comment, @rubanm ?

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  private lazy val db = factory.open(dir, options)
```
johnynek: is this thread-safe? Can you link to docs showing this? If not, we should use:
https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/concurrent/AsyncMutex.scala

Then we'd do something like:
```scala
dbMutex.acquireAndRun(futurePool {
  // method logic here.
})
```

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
rubanm: Makes sense to me.

@BenFradet Could you change to something like:
```scala
class LevelDBStore[K: Codec, V: Codec]( ... )
  extends Store[K, V]
```
There are a few existing stores that use Codec to convert between T and Array[Byte] that can serve as examples.

ianoc: It can be a performance concern, but unless its been identified as one I'd just leave it to combinators

rubanm: Ah, sorry I misread the comment. Yes combinators it is then. @BenFradet please disregard my earlier comment.

BenFradet:
Thanks for the feedback, will take it into account.

storehaus-leveldb/src/test/scala/com/twitter/storehaus/leveldb/LevelDBStoreProperties.scala
BenFradet: Yup, I meant to change that but forgot, thanks for reminding me.

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
BenFradet: Will do.

storehaus-leveldb/src/test/scala/com/twitter/storehaus/leveldb/LevelDBStoreProperties.scala
BenFradet: Yup.

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
BenFradet: Will look into it and post my findings.

project/Build.scala
```diff
@@ -226,19 +226,29 @@ object StorehausBuild extends Build {
...
 
+  lazy val storehausLevelDB = module("leveldb").settings(
+    libraryDependencies +=
+      "org.fusesource.leveldbjni" % "leveldbjni-all" % "1.8",
+    parallelExecution in Test := false,
+    // workaround because of how sbt handles native libraries
+    // http://stackoverflow.com/questions/19425613/unsatisfiedlinkerror-with-native-library-under-sbt
+    testOptions in Test := Seq(),
+    fork in Test := true
+  ).dependsOn(storehausAlgebra % "test->test;compile->compile")
```
rubanm: This can be just `storehausCore`?

rubanm:
LGTM! @johnynek / @ianoc could one of you take another look and merge if all good?

BenFradet:
I haven't looked into whether

```scala
private lazy val db = factory.open(dir, options)
```

was thread-safe or not yet.

project/Build.scala
BenFradet: Yup, I think that's what I meant to do when I did [this](https://github.com/twitter/storehaus/pull/258/files#diff-215113124f1de02f228327ba7abb45f7R251).

BenFradet:
Indeed, it's not thread-safe (cf [the code](https://github.com/fusesource/leveldbjni/blob/master/leveldbjni/src/main/java/org/fusesource/leveldbjni/JniDBFactory.java#L163), I'll follow @johnynek's advice and use a mutex.

BenFradet:
My bad, it seems to be thread-safe after all (cf [this](https://github.com/dain/leveldb/blob/master/leveldb/src/main/java/org/iq80/leveldb/impl/DbImpl.java)).

storehaus-leveldb/src/main/scala/com.twitter.storehaus.leveldb/LevelDBStore.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+    kvs.mapValues(_ => future)
+  }
+
+  /**
+   * Close this store and release any resources.
+   * It is undefined what happens on get/multiGet after close
+   */
+  override def close(time: Time): Future[Unit] = {
+    futurePool { db.close() }
```
johnynek: this should be: `futurePool { db.close() }.flatMap { _ => super.close(time) }`

johnynek:
one minor issue, and remerge with develop and it should be good. Thanks.

BenFradet:
Will do.

BenFradet:
Travis fails due to an ElasticSearch test.

rubanm:
Restarted the build.

BenFradet:
I'll merge develop in order to take advantage of #267 soon.

rubanm:
@BenFradet Sorry, looks like we missed this one for the ongoing minor version release. Yes let's merge once this is green and we can add it to the next release.

BenFradet:
@rubanm yup, no problem. Hopefully, it'll turn green this time.

rubanm:
Merging, thanks!

PR #261, opened 2015-05-05
Remove usage of twitter's maven repo, travis seems to dislike it -- mayb...

ianoc:
...e routing issues?

Also eliminated the finagle module code, didn't seem to add anything since we don't need to cross fancily now.


johnynek:
@mosesn what's up with our stuff not being on maven central? Is that just old stuff or are we really not publishing some things?

travisbrown:
@johnynek There are two major reasons that `maven.twttr.com` is still around. The first is that we're publishing artifacts into group IDs we don't own (like our ancient fork of `libthrift`). We could change the group ID on these, but the other problem is that we don't have a way to publish lots of project (like commons) to Maven Central, since they only have Pants builds.

We have internal issues open for updating libthrift and we've discussed putting together a SBT build for commons, but it hasn't been a priority until very recently, when an internal change to the `maven.twttr.com` servers meant that we started getting weird intermittent errors. I can send you a link to that issue as well, and I've got a couple of workarounds, including a new one (just adding `-Dhttp.keepAlive=false` to the SBT launcher) that seems to fix it entirely (but is still a pain in the ass).

mosesn:
Pants also can't publish to maven central afaik.

travisbrown:
@mosesn Yeah, sorry if that wasn't clear from my first paragraph.

johnynek:
It seems pants publishing to oss.sonatype would be a big help.

So, from our perspective, I guess we don't need the maven.twttr, so we can drop it safely. And, longer term, it looks like it is not going anywhere any time sonn for at least two reasons. Right?

PR #262, opened 2015-05-12
storehaus-memcache: pass ttl for MergeableMemcacheStore CAS calls

franklinhu:
MergeableMemcacheStore was not passing the right TTL setting, and thus was using the default of Time.epoch which never expires the item.

cc @danielhfrank

### DIFF
--- a/storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MergeableMemcacheStore.scala
+++ b/storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MergeableMemcacheStore.scala
@@ -77,7 +77,7 @@ class MergeableMemcacheStore[K, V](underlying: MemcacheStore, maxRetries: Int)(k
               inj.invert(cbValue) match {
                 case Success(v) => // attempt cas
                   val resV = semigroup.plus(v, kv._2)
-                  underlying.client.cas(key, 0, underlying.ttl.fromNow, inj.apply(resV), casunique).flatMap { success =>
+                  underlying.client.cas(key, underlying.flag, underlying.ttl.fromNow, inj.apply(resV), casunique).flatMap { success =>
                     success.booleanValue match {
                       case true => Future.value(Some(v))
                       case false => doMerge(kv, currentRetry + 1) // retry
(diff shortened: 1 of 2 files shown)


danielhfrank:
@franklinhu we'll want to make the same adjustment to the call to `add` below

franklinhu:
@danielhfrank Thanks, fixed in latest rev

rubanm:
Thanks for the fix! Will merge when green.

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MergeableMemcacheStore.scala
```diff
@@ -77,7 +77,7 @@ class MergeableMemcacheStore[K, V](underlying: MemcacheStore, maxRetries: Int)(k
               inj.invert(cbValue) match {
                 case Success(v) => // attempt cas
                   val resV = semigroup.plus(v, kv._2)
-                  underlying.client.cas(key, inj.apply(resV), casunique).flatMap { success =>
+                  underlying.client.cas(key, 0, underlying.ttl.fromNow, inj.apply(resV), casunique).flatMap { success =>
```
rubanm: Actually, the second arg should be `underlying.flag` instead of 0?

franklinhu: I have no idea what flag does, but this would change the existing behavior right? Should we also pass flags into CAS?

rubanm: Found an explanation here
https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L155
and here
http://stackoverflow.com/questions/12228097/what-is-the-flags-arg-in-finagle-memcached-set-method

The flag is opaque to the server and is returned back along with the value when reading. For correctness, it looks like we should set the same flag as the underlying store does for `set` calls. It's probably been working fine all this while because of the default being 0 in `MemcacheStore`. Not your change, but would be nice to fix this as well. What do you think?

rubanm: Ah, didn't notice your latest rev. Should be added to the cas call as well I think.

franklinhu:
Fixed the review feedback. Looks like the test failures are unrelated to the memcache changes

johnynek:
restarted the build.

rubanm:
Green now. Merging, thanks.

PR #266, opened 2015-05-19
Ugprade finagle-memcached to finagle-memcachedx

luciferous:


storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheStore.scala
```diff
@@ -119,23 +120,27 @@ class MemcacheStore(val client: Client, val ttl: Duration, val flag: Int)
 {
   override def withPutTtl(ttl: Duration) = new MemcacheStore(client, ttl, flag)
 
-  override def get(k: String): Future[Option[ChannelBuffer]] = client.get(k)
+  override def get(k: String): Future[Option[ChannelBuffer]] =
+    client.get(k).map(_.map(ChannelBufferBuf.Owned.extract))
```
luciferous: Would you like if I pattern matched on the Option here to avoid the allocation for None case (using Future.None)?

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MergeableMemcacheStore.scala
```diff
@@ -71,13 +73,20 @@ class MergeableMemcacheStore[K, V](underlying: MemcacheStore, maxRetries: Int)(k
     val key = kfn(kv._1)
     (currentRetry > maxRetries) match {
       case false => // use 'gets' api to obtain casunique token
-        underlying.client.gets(key).flatMap { res : Option[(ChannelBuffer, ChannelBuffer)] =>
+        underlying.client.gets(key).flatMap { res =>
           res match {
```
luciferous: Would you mind if I flattened this?

```scala
underlying.client.gets(key).flatMap {
  case Some((cbValue, casunique)) => // ...
  case None => // ...
}
```

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheStore.scala
luciferous: If it's not hot, I'd prefer to just keep it as is.

luciferous:
Reran the two failures observed in Travis build, both succeeded.

```
$ ./sbt 'project storehaus-cache' 'testOnly com.twitter.storehaus.cache.MutableTTLCacheProperties'
[...snip...]
[info] + MutableTTLCache.If insert an item with a TTL of X, and wait X + delta then it cannot be there: OK, passed 100 tests.
[info] + MutableTTLCache.if you put with TTL t, and wait 0 time, it should always (almost, except due to scheduling) be in there: OK, passed 100 tests.
[...snip...]
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
```

```
$ ./sbt 'project storehaus-elasticsearch' 'testOnly com.twitter.storehaus.elasticsearch.ElasticSearchStoreSpecs'
[...snip...]
[info] ElasticSearchStoreSpecs:
[info] ElasticSearch Store
[info] - should Put a value
[info] - should Retrieve a value that doesnt exist
[info] - should Update a value
[info] - should Delete a value
[info] - should Put multiple values
[info] - should Retrieve values that do not exist
[info] - should Update multiple values
[info] - should Delete multiple values
[info] - should Search for  values
[info] ScalaTest
[info] Run completed in 42 seconds, 474 milliseconds.
[info] Total number of tests run: 9
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 9, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 9, Failed 0, Errors 0, Passed 9
```

luciferous:
Would you like me to squash the commits before merge?

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MergeableMemcacheStore.scala
rubanm: sounds good

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheStore.scala
rubanm: I don't think we've seen any regression because of this, but @ianoc might know more. Doesn't hurt to change it I guess.

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheLongStore.scala
```diff
@@ -34,9 +35,16 @@ import com.twitter.bijection.netty.ChannelBufferBijection
  */
 object MemcacheLongStore {
   private implicit val cb2ary = ChannelBufferBijection
+  private implicit val b2ary = new Bijection[Buf, Array[Byte]] {
```
rubanm: We have `AbstractBijection` for this. A little clearer than extending the trait.

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheLongStore.scala
```diff
@@ -34,9 +35,16 @@ import com.twitter.bijection.netty.ChannelBufferBijection
...
+    def apply(buf: Buf) = Buf.ByteArray.Owned.extract(buf)
```
rubanm: `override def`

luciferous: It doesn't override though.

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheLongStore.scala
luciferous: From [Bijection.scala](https://github.com/twitter/bijection/blob/develop/bijection-core/src/main/scala/com/twitter/bijection/Bijection.scala#L68):
```scala
/**
 * Abstract class to ease Bijection creation from Java.
 */
abstract class AbstractBijection[A, B] extends Bijection[A, B] {
```

Doesn't this suggest we should be implementing Bijection directly?

rubanm: It was originally created for java compatibility yes but we have mixed usage of that in scala as well. nbd though, this can be left as is.

storehaus-memcache/src/main/scala/com/twitter/storehaus/memcache/MemcacheLongStore.scala
rubanm: This is more of a variation in style I guess. Personally, override is clearer to the reader without having to look if it comes from the base trait. Maybe doesn't make sense to do this for `apply`.

luciferous: I changed this to use the builtin Buf bijection which I just realized we have. This is now moot.

rubanm:
LGTM. Merge when green (or if the two flaky tests pass locally).

rubanm:
This is green now.. merging

Issue #270, opened 2015-06-02
Wrong spelling in AsyncHBaseStringStore.apply


yaochitc:
Should the returned instance of the apply method be AsyncHBaseStringStore instead of AsyncHBaseLongStore?

object AsyncHBaseStringStore {
  def apply(quorumNames: Seq[String],
            table: String,
            columnFamily: String,
            column: String,
            threads: Int=4): AsyncHBaseLongStore = {
    val store = new AsyncHBaseLongStore(quorumNames, table, columnFamily, column, new HBaseClient(quorumNames.mkString(",")), threads)
    store.validateConfiguration()
    store
  }
}

MansurAshraf:
Thanks, created a Patch with the fix

PR #271, opened 2015-06-02
AsyncHBaseStringStore is returning wrong instance

MansurAshraf:
fixes https://github.com/twitter/storehaus/issues/270


johnynek:
While you are at it, can you just directly call the hbase string injection?

https://github.com/twitter/storehaus/pull/271/files#diff-0081b294174d955d5f2bc1937683188fR51

Rather than using implicit resolution for a concrete type. It would make the code a bit easier to follow.

PR #274, opened 2015-07-28
Updates for 2.11

ianoc:
Had to drop storehaus-kafka, seemed to manually reference kafka_2.9.2 . So I doubt it even worked for 2.10 properly. Doesn't at all for 2.11 . We have kafka 0.8 for 2.10 and 2.11.

Also removed all usages of specs2, using scalatest where it was so we can be more consistent


rubanm:
LGTM

PR #279, opened 2015-08-24
Remove dependency on finagle-http

luciferous:
Finagle is deprecating the http module in favor of httpx (a similar
module but without leaking underlying implementation i.e. Netty types).

This patch doesn't remove Netty but preserves storehaus's
Netty-dependent API while removing the finagle-http dependency.
(Storehaus can choose to deprecate the Netty API and adopt the
finagle-httpx interface at its own pace.)


johnynek:
restarting CI. It *looks* like the 2.11 had some flakey test.

PR #282, opened 2015-09-07
Feature/storehaus cassandra only

ghost:
As discussed in #211


project/Build.scala
```diff
@@ -31,6 +31,17 @@ object StorehausBuild extends Build {
...
     }
+
+  def bogus210Version(dep: ModuleID) = dep cross CrossVersion.binaryMapped {
+    case version if version startsWith "2.10" => "2.10.3"
+    case x => x
+  }
+
+  def isScalaGreater29x(scalaVersion: String) = scalaVersion match {
+    case version if version startsWith "2.9" => false
+    case version if version matches "^2.1[0-2].*" => true
```
johnynek: 1) I think you want: `^2\.1[0-2]\.*` here.
2) Can we use the same approach in all the new code we add? startsWith is dangerous (consider version 2.91.0 which is not the same as 2.9.x).

project/Build.scala
```diff
@@ -60,6 +71,12 @@ object StorehausBuild extends Build {
       Opts.resolver.sonatypeSnapshots,
       Opts.resolver.sonatypeReleases,
       "Conjars Repository" at "http://conjars.org/repo"
+//      "Bintray JCenter" at "https://jcenter.bintray.com/",
```
johnynek: please don't add commented code.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
```diff
@@ -0,0 +1,126 @@
...
+    	    }
+    	    mutator.add(builder)
+    	  }
+    	  case (key, None) => mutator.add(deleteColumns match {
+    	    case None => QueryBuilder.delete().from(columnFamily.getName).where(QueryBuilder.eq(keyColumnName, key))
+    	    case Some(cols) => QueryBuilder.delete(cols).from(columnFamily.getName).where(QueryBuilder.eq(keyColumnName, key))
+    	  })
+    	}
+    	mutator.setConsistencyLevel(consistency)
+    	columnFamily.session.getSession.execute(mutator)
```
johnynek: is this thread-safe? Should we use the AsyncMutex to ensure only one query is running?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
```diff
@@ -0,0 +1,126 @@
...
+   */
+  override def multiGet[K1 <: K](ks: Set[K1]): Map[K1, Future[Option[V]]] = {
+    import scala.math.Ordering.Implicits._
+    import scala.collection.convert.WrapAsScala._
+    cassandraVersion match {
+      case Some(version) if (version.getMajor, version.getMinor) >= ((2, 1)) =>
+        val clause = QueryBuilder.in(keyColumnName, ks.map(k => keySerializer.toCType(k)))
+        val stmt = QueryBuilder.select().from(columnFamily.getPreparedNamed).where(clause).setConsistencyLevel(consistency)
+        val future = futurePool { 
+          val result = columnFamily.session.getSession.execute(stmt)
```
johnynek: can we add a comment about why this is thread-safe here?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
```diff
@@ -0,0 +1,126 @@
...
+        .from(columnFamily.getPreparedNamed)
+        .where(QueryBuilder.eq(keyColumnName, key))
+  }
+  
+  override def get(key: K): Future[Option[V]] = {
+    futurePool {
+      val stmt = createGetQuery(key)
+        .limit(1)
+        .setConsistencyLevel(consistency)
+      val result = columnFamily.session.getSession.execute(stmt)
```
johnynek: thread safety question here.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraStore.scala
```diff
@@ -0,0 +1,106 @@
...
+  		Some(iterableAsScalaIterableConverter(rSet).asScala.view.toSeq.map(row => getKeyValueFromRow(row)))
+  	  }
+    }
+  }
+  
+  /**
+   * allows iteration over all (K, V) using a query against queryable
+   */
+  override def getAll: Future[Spool[(K, V)]] = queryable.get("").transform {
+    case Throw(y) => Future.exception(y)
```
johnynek: isn't this flatMap? Why transform here?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/CASStore.scala
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * represents store supporting Compare/Check and Set. This allows to implement 
+ * distributed asynchronous "transaction-like" operations. E.g. a distributed 
+ * lock implemented with this and Futures is a snap.
+ */
+trait CASStore[T, K, V] {
+  
+  def get(k: K)(implicit ev1: Equiv[T]): Future[Option[(V, T)]]
+  
+  def cas(token: Option[T], kv: (K, V))(implicit ev1: Equiv[T]): Future[Boolean]
```
johnynek: why does this take an `Option[T]` rather than T? same question below.

johnynek:
@travisbrown can you review the shapeless usage here? I know very little shapeless.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/CQLCassandraStoreTupleValues.scala
```diff
@@ -0,0 +1,173 @@
...
+import com.twitter.storehaus.{IterableStore, QueryableStore, ReadableStore, ReadableStoreProxy, Store, WithPutTtl}
+import com.twitter.util.{Await, Closable, Future, Duration, FuturePool, Promise, Try, Throw, Return}
+import com.websudos.phantom.CassandraPrimitive
+import java.util.concurrent.Executors
+import shapeless._
+import shapeless.ops.hlist.{Mapper, ToList, Tupler}
+import shapeless.ops.traversable.FromTraversable
+import shapeless.UnaryTCConstraint.*->*
+import shapeless.syntax.std.product._
+import syntax.std.traversable._
```
travisbrown: `shapeless.syntax.std.traversable._` would be better for consistency and clarity.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/macrosbug/hListerAux.scala
```diff
@@ -0,0 +1,128 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.twitter.storehaus.cassandra.cql.macrobug
+
+import shapeless._
+
+trait HListerAux[-T <: Product, Out <: HList] { def apply(t : T) : Out }
```
travisbrown: Why not `Generic`?

project/Build.scala
```diff
@@ -293,6 +316,26 @@ object StorehausBuild extends Build {
...
   ).dependsOn(storehausCore, storehausAlgebra, storehausCache)
 
+  def cassandraDeps(scalaVersion: String) = if (!isScalaGreater29x(scalaVersion)) Seq() else Seq(
+    "com.twitter" %% "algebird-core" % algebirdVersion,
+    "com.google.code.findbugs" % "jsr305" % "1.3.+",
+    "com.twitter" %% "bijection-core" % bijectionVersion,
+    "com.datastax.cassandra" % "cassandra-driver-core" % cassandraDriverVersion classifier "shaded" exclude("io.netty", "*"),
+    "com.websudos" %% "phantom-dsl" % "1.5.0" exclude ("com.datastax.cassandra", "cassandra-driver-core"),
+    withCross("com.twitter" %% "util-zk" % utilVersion)  exclude ("com.google.guava", "guava"),
+    bogus210Version("com.chuusai" %% "shapeless" % "2.0.0"),
```
travisbrown: Why not 2.2.5? Tests pass just fine on 2.2.5 with a couple of minor changes (that are arguably improvements, anyway).

project/Build.scala
```diff
@@ -122,6 +139,11 @@ object StorehausBuild extends Build {
   val scaldingVersion = "0.15.0"
   val finagleVersion = "6.27.0"
   val scalatestVersion = "2.2.4"
+  val cascadingVersion = "2.5.2"
+  val hadoopVersion = "1.2.1"
```
ianoc: why hadoop1, not hadoop 2?

project/Build.scala
```diff
@@ -31,6 +31,17 @@ object StorehausBuild extends Build {
       case version if version startsWith "2.10" => "2.10" // TODO: hack because sbt is broken
       case x => x
     }
+
+  def bogus210Version(dep: ModuleID) = dep cross CrossVersion.binaryMapped {
+    case version if version startsWith "2.10" => "2.10.3"
+    case x => x
+  }
+
+  def isScalaGreater29x(scalaVersion: String) = scalaVersion match {
```
ianoc: We don't cross compile for 2.9 anymore, do we need this at all?

project/Build.scala
```diff
@@ -293,6 +316,26 @@ object StorehausBuild extends Build {
     javaOptions in run <++= (fullClasspath in Runtime) map { cp => Seq("-cp", sbt.Build.data(cp).mkString(":")) }
   ).dependsOn(storehausCore, storehausAlgebra, storehausCache)
 
+  def cassandraDeps(scalaVersion: String) = if (!isScalaGreater29x(scalaVersion)) Seq() else Seq(
+    "com.twitter" %% "algebird-core" % algebirdVersion,
+    "com.google.code.findbugs" % "jsr305" % "1.3.+",
+    "com.twitter" %% "bijection-core" % bijectionVersion,
+    "com.datastax.cassandra" % "cassandra-driver-core" % cassandraDriverVersion classifier "shaded" exclude("io.netty", "*"),
```
ianoc: can we depend on cassandra core normally and just mark it as provided?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/CQLCassandraMultivalueStore.scala
```diff
@@ -0,0 +1,162 @@
...
+  ttl: Option[Duration] = CQLCassandraConfiguration.DEFAULT_TTL_DURATION)
+  (cassSerValue: VS, valueColumnNameList: List[String])(
+    implicit evrow: Mapped.Aux[RK, CassandraPrimitive, RS],
+    evcol: Mapped.Aux[CK, CassandraPrimitive, CS],
+    evval: Mapped.Aux[V, CassandraPrimitive, VS],
+    rowmap: AbstractCQLCassandraCompositeStore.Row2Result[RK, RS],
+    colmap: AbstractCQLCassandraCompositeStore.Row2Result[CK, CS],
+    a2cRow: AbstractCQLCassandraCompositeStore.Append2Composite[ArrayBuffer[Clause], RK, RS], 
+    a2cCol: AbstractCQLCassandraCompositeStore.Append2Composite[ArrayBuffer[Clause], CK, CS],
+    rsUTC: *->*[CassandraPrimitive]#λ[RS],
```
travisbrown: These `*->*` are strictly weaker than their `Mapped` counterparts above.

ianoc:
I think I'd rather we make a storehaus-contrib's repo and move these sorts of stores there. Any upgrades to support new scala versions to the core get blocked on lots of these external stores with questionable external support around them. I'd rather we have a core which we can be very confident about and the contrib repo people can update or not update as they like. I'm afraid of our builds being blocked by say flaky tests in elastic search that we know nothing about the code. Thoughts @johnynek ?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+   */
+  object keyStringMapping extends (CassandraPrimitive ~>> String) {
+    override def apply[T](c: CassandraPrimitive[T]) = c.cassandraType
+  }
+  
+  /**
+   * used to map over an example key to implicit serializers
+   */
+  object cassandraSerializerCreation extends Poly1 {
+    implicit def default[T : CassandraPrimitive] = at[T](_ => implicitly[CassandraPrimitive[T]])
```
travisbrown: I'd recommend including the `Case.Aux` annotation here.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  /**
+   * creates an HList of serializers (CasasandraPrimitives) given an example 
+   * key HList and pulls serializers out of implicit scope.
+   */
+  def getSerializerHListByExample[KL <: HList, SL <: HList](keys: KL)(implicit mapper: Mapper.Aux[cassandraSerializerCreation.type, KL, SL]) : SL = {
```
travisbrown: You shouldn't need the `Aux` here—using `mapper.Out` as the return type should give you the same static type.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+import com.datastax.driver.core.querybuilder.{BuiltStatement, Clause, QueryBuilder, Update, Select}
+import com.twitter.storehaus.{IterableStore, QueryableStore, ReadableStore, Store}
+import com.websudos.phantom.CassandraPrimitive
+import java.util.concurrent.{Executors, TimeUnit}
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ArrayBuffer
+import scala.language.implicitConversions
+import shapeless._
+import HList._
```
travisbrown: This import looks unnecessary?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+import ops.hlist.Mapper
+import ops.hlist.Mapped
+import poly._
+import Nat._
```
travisbrown: Also unnecessary?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+import java.util.concurrent.{Executors, TimeUnit}
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ArrayBuffer
+import scala.language.implicitConversions
+import shapeless._
+import HList._
+import ops.hlist.Mapper
+import ops.hlist.Mapped
+import poly._
```
travisbrown: I'm recommend specifically importing `~>>` for clarity.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+import Nat._
+import UnaryTCConstraint._
```
travisbrown: 99% of the times I think I have a use for `*->*` it turns out that I don't, and I don't think you need it here (see inline comments below).

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+   * helper implicits for the recursion itself
+   */
+  object Append2Composite {
+    implicit def hnilAppend2Composite[R <: ArrayBuffer[Clause]] = new Append2Composite[R, HNil, HNil] {
+      override def apply(r: R, k: HNil, s: List[String], ser: HNil): Unit = {}
+    }
+    implicit def hlistAppend2Composite[R <: ArrayBuffer[Clause], M, K <: HList, P, Q <: HList](
+      implicit a2c: Append2Composite[R, K, Q]) = new Append2Composite[R, M :: K, P :: Q] {
+    	override def apply(r: R, k: M :: K, s: List[String], ser: P :: Q): Unit = {
+   	      r.add(QueryBuilder.eq(s"""\"${s.head}\"""", ser.head.asInstanceOf[CassandraPrimitive[Any]].toCType(k.head)))
```
travisbrown: If you know that `P` is a `CassandraPrimitive` why not capture that in the types?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+import Nat._
+import UnaryTCConstraint._
+import org.slf4j.LoggerFactory
+
+object AbstractCQLCassandraCompositeStore {
+  
+  /**
+   * used for mapping the keys in the HLists to Strings
+   */
+  object keyStringMapping extends (CassandraPrimitive ~>> String) {
```
travisbrown: I'd be very tempted to define a custom type class that provides a conversion directly from an `HList` of `CassandraPrimitive`s to a `List[String]`. It would save a lot of `Mapper` / `ToList` pairs below, and I personally think I'd find it clearer.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/cassandraTupleStores.scala
```diff
@@ -0,0 +1,128 @@
...
+import shapeless._
+import com.twitter.storehaus.cassandra.cql.macrobug._
+
+/**
+ * AbstractCQLCassandraCompositeStore-wrapper for Tuples to HList. This makes it 
+ * easy to use this store with tuples, which can in turn be serialized easily.
+ */
+class CassandraTupleStore[RKT <: Product, CKT <: Product, V, RK <: HList, CK <: HList, RS <: HList, CS <: HList]
+		(val store: AbstractCQLCassandraCompositeStore[RK, CK, V, RS, CS], paramToPreventWritingDownTypes: (RKT, CKT))
+		(implicit ev1: HListerAux[RKT, RK],
```
travisbrown: Using `Generic` would give you both directions in a single type class (and allow you to avoid all the `HListerAux` boilerplate). If you want evidence that you've got a tuple (as opposed to some other `Product`, there's an `IsTuple` type class).

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+   */
+  trait Row2Result[K <: HList, S <: HList] {
+    def apply(row: Row, serializers: S, names: List[String]): K
+  }
+
+  /**
+   * helper implicits for the recursion itself
+   */
+  object Row2Result {
+    implicit def hnilRow2Result = new Row2Result[HNil, HNil] {
```
travisbrown: Type annotations would be better.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+      override def apply(row: Row, serializers: HNil, names: List[String]): HNil = HNil
+    }
+    implicit def hlistRow2Result[M, K <: HList, N, S <: HList](
+      implicit r2r: Row2Result[K, S]) = new Row2Result[M :: K, N :: S] {
+    	override def apply(row: Row, serializers: N :: S, names: List[String]): M :: K = {
+    	  val ser = serializers.head.asInstanceOf[CassandraPrimitive[M]]
```
travisbrown: Why not skip the `N` and just use `CassandraPrimitive[M]` directly?

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
```diff
@@ -0,0 +1,294 @@
...
+   */
+  implicit def append2Composite[R <: ArrayBuffer[Clause], K <: HList, S <: List[String], Q <: HList](r: R)(k: K, s: List[String], ser: Q)
+  	(implicit a2c: Append2Composite[R, K, Q]) = a2c(r, k, s, ser) 
+    
+  /**
+   * helper trait for declaring the HList recursive function 
+   * to create a key from a row
+   * K is the type of HList to return the key, S are the serializers
+   */
+  trait Row2Result[K <: HList, S <: HList] {
```
travisbrown: This is a candidate for `DepFn1` treatment, since `K` uniquely determines the `S`.

project/Build.scala
Rechzen: I will do as ianoc suggested and throw that out.

project/Build.scala
AndreasPetter: Yes... sorry.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
AndreasPetter: It is thread safe. I'll add http://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/Session.html as a comment

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
AndreasPetter: same here

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraSimpleStore.scala
AndreasPetter: same here

project/Build.scala
AndreasPetter: 2.2.5 doesn't seem to distinguish between scala 2.10.x any more. is it safe to assume that it will be working with all 2.10s?

project/Build.scala
AndreasPetter: sorry, must be removed as this pr does not depend on cascading nor hadoop

project/Build.scala
AndreasPetter: no, thanks, will remove it

project/Build.scala
travisbrown: @AndreasPetter It'll work with any 2.10 after 2.10.1.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
AndreasPetter: Sorry to bother you again, but I'm not able to bring "annotation" in useful context here - I'm not really a shapeless pro. What I do understand is that I use CaseBuilder (and therefore Case) under the hood by using `at`. I found an annotation.scala in shapeless but as far as I understand (and i must admit that i did not understand too much there) this is for witnessing annotations on types - for which I do not know how to use them here. Maybe you meant that it would be beneficial to add the Case.Aux implicit like `implicit c: Case.Aux[T, CassandraPrimtive[T]]`?

AndreasPetter: @travisbrown ah sorry, I just understood that you stated that it would be better to specify the return type with Case.Aux. Forget my previous comment, sorry.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/AbstractCQLCassandraCompositeStore.scala
AndreasPetter: Sorry to bother you again... I made a few changes such that `trait Row2Result[S <: HList] extends DepFn1` creating some dependent `Aux[S <: HList, K <: HList]` and using the dependent `Out` type from `DepFn1`. It (almost) seems to work except that the compiler complains that the apply method with 1 parameter is not available and object creation would be impossible without it. I succeeded creating a special `DepFn1ApplyParams3` having an apply method that takes the number of parameters I'm using - skipping the original `DepFn1`, however. Is this the way to go, or is there a better way to do it?

travisbrown: Sorry, I was using `DepFn1` as shorthand for the approach of making the uniquely-determined type parameter a type member. If the `apply` on `DepFn1` doesn't have the right signature, it'd make sense just to change this to `RowResult[K <: HList] { type Out <: HList ...`.

storehaus-cassandra/src/main/scala/com/twitter/storehaus/cassandra/cql/CASStore.scala
AndreasPetter: The semantics of CASStore was indeed already subject to an internal emotional debate a few month ago. Added a comment and some tests showing the semantics which is implemented: None means, that there must not exist a token previously (and the operation is only performed, if that is the case) and the store sets the first one using some token-factory; while providing Some(token) means that the store needs to check for that token and a new one is being set during the operation iff the previously existing token is stored before/during the start of the operation. One could - of course - define a variety of other semantics.

AndreasPetter:
Hm... Travis tests are failing, with the note "Killed". Since they run locally and no error shows up it is not unlikely that this has something to do with https://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error . How do you arrange tests at Twitter if this happens?

AndreasPetter:
O.k. I had to use the shaded cassandra driver (sorry @ianoc, there seem to be dependency conflicts with the driver with the other stores), and use an embedded Cassandra version for testing. Are there any more comments?

PR #284, opened 2016-02-02
Update the build to supprt sbt-release

johnynek:


ianoc:
Removed the not-needed Scrooge plugin (i'm guessing copied from another repo?). Doesn't seem like it was able to easily download it, those versions of apache are ancient non-mainline ones i thinks scrooge uses. So just happy other places are working still... :)

PR #286, opened 2016-03-05
Kafka cleanup

BenFradet:
This PR aims to cleanup the kafka module:
- upgrade kafka to 0.9.0.1
- usage of the new producer
- rename the module to storehaus-kafka (previously storehaus-kafka08)

I'm also planning on making the associated specs real unit tests and not integration tests (i.e. remove the dependency on an external kafka broker) if there is interest.

Review welcome!


storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaInjections.scala
```diff
@@ -16,40 +16,48 @@
...
+  trait FromInjectionEncoder[T] extends Serializer[T] with SerDe {
     def injection: Injection[T, Array[Byte]]
 
-    override def toBytes(t: T): Array[Byte] = injection(t)
+    override def serialize(topic: String, data: T): Array[Byte] =
+      injection(data)
   }
 
-  def fromInjection[T: Codec]: (Encoder[T], Decoder[T]) = {
+  trait SerDe {
```
johnynek, line 47: can you document this class. Clearly mutation is going on behind the scenes. We prefer to avoid that when at all possible.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,102 @@
...
+  private lazy val producer = new KafkaProducer[K, V](props)
+
+  /**
+    * Put a key/value pair in a Kafka topic
+    * @param kv (key, value)
+    * @return Future.unit
+    */
+  override def put(kv: (K, V)): Future[Unit] = Future {
+    val (key, value) = kv
+    producer.send(new ProducerRecord[K, V](topic, key, value))
```
johnynek: what is the threading story here? Is this async?

Is it threadsafe?

The `Future { }` as you probably know only catches exceptions but does the work in this thread.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /**
+    * Put multiple (key, value) pairs at one time
+    * @param kvs (key, value) pairs
+    */
+  override def multiPut[K1 <: K](kvs: Map[K1, V]): Map[K1, Future[Unit]] = {
```
johnynek: No need to implement this if there is nothing faster than a bunch of `put` calls. This is here for systems that have batch APIs.

Also, this is imprecise because if all but the last succeed, the caller will think they all failed. At least fix that issue.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,102 @@
...
+    * @param brokers Addresses of the Kafka brokers in the hostname:port format
+    * @return Kafka Store
+    */
+  def apply[K, V, KS <: Serializer[K] : Manifest, VS <: Serializer[V] : Manifest](
+    topic: String,
+    brokers: Seq[String]
+  ) = new KafkaStore[K, V](topic, createProps[K, V, KS, VS](brokers))
+
+
+  private def createProps[K, V, KS <: Serializer[K] : Manifest, VS <: Serializer[V] : Manifest](
```
johnynek: can we use `ClassTag` here, which is what replaced Manifest?

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaInjections.scala
```diff
@@ -16,40 +16,48 @@
...
+    def configure(configs: util.Map[String, _], isKey: Boolean): Unit = ()
+    def close(): Unit = ()
+  }
+
+  def fromInjection[T: Codec]: (Serializer[T], Deserializer[T]) = {
```
johnynek, line 52: how do these get used if below? It seems kafka creates the serializers with reflection.

build.sbt
```diff
@@ -237,11 +236,11 @@ lazy val storehausLevelDB = module("leveldb").settings(
   fork in Test := true
 ).dependsOn(storehausCore % "test->test;compile->compile")
 
-lazy val storehausKafka08 = module("kafka-08").settings(
```
johnynek, line 240: this is going to change the jar name.

What do you think of having a kafka-09 module instead of ending support for 08? I actually don't know if anyone is using the 08 modules, but I hate to force an upgrade on someone or more realistically leave them stranded on an old version of storehaus.

BenFradet, line 240: The main goal was to move to the new producer which is supported in 0.8 and 0.9 and having only one jar which would be named `storehaus-kafka` instead of having one per version.
What do you think?

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaInjections.scala
BenFradet, line 52: AFAIK, this all object is not used anywhere and was thinking of removing it.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
BenFradet: This is async and send back a java future but I wouldn't know how to convert it to a twitter-util future.
If you have any insight on how to do that, that'd be greatly appreciated.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaInjections.scala
MansurAshraf, line 52: Its used [here](https://github.com/twitter/storehaus/blob/develop/storehaus-kafka-08/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala#L92). Lets not remove it as its very convenient to use Injections as Kafka Encoders

BenFradet:
I made the related specs unit tests and not integration tests and addressed some of your comments @johnynek

BenFradet:
Pinging @johnynek to get your input regarding the artifact naming / split.

johnynek:
sorry for letting this get lost.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaInjections.scala
johnynek, line 52: yeah, I see that Kafka has non-reflective ways to pass the Serializers, so let's leave them.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaAvroSink.scala
```diff
@@ -0,0 +1,93 @@
...
+    * topic
+    * @param topic Kafka topic
+    * @param brokers Addresses of the Kafka brokers in the hostname:port format
+    * @tparam V Type of the Avro record
+    * @return KafkaSink[String, SpecificRecordBase]
+    */
+  def apply[V <: SpecificRecordBase : Manifest](
+    topic: String,
+    brokers: Seq[String],
+    executor: => ExecutorService
```
johnynek: I don't see where this is used.

BenFradet: Yup, it seems to be an artifact from the last implementation.
Will remove.

BenFradet:
No problem, thanks for reviewing.

BenFradet:
The tests seem to be a bit flaky, I'll investigate.

BenFradet:
The failing test doesn't seem to be related to my changes:

> [error] Failed tests:
> [error]     com.twitter.storehaus.cache.MutableTTLCacheProperties

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaAvroSink.scala
```diff
@@ -0,0 +1,92 @@
...
+    */
+  def apply[V <: SpecificRecordBase : Manifest](
+    topic: String,
+    brokers: Seq[String]
+  ): KafkaSink[String, V] = {
+    implicit val inj = SpecificAvroCodecs[V]
+    lazy val sink =
+      KafkaSink[Array[Byte], Array[Byte], ByteArraySerializer, ByteArraySerializer](
+        topic: String, brokers: Seq[String]).convert[String, V](utf8.toFunction)
+    sink
```
johnynek: sorry, just noticed this pattern. What is going on here? Why lazy, then we force it? Why not just return line 51 directly?

Same below...

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaSink.scala
```diff
@@ -78,40 +74,36 @@ object KafkaSink {
...
+    * @param brokers Addresses of the Kafka brokers in the hostname:port format
+    * @return KafkaSink[K, V]
+    */
+  def apply[K, V, KS <: Serializer[K] : Manifest, VS <: Serializer[V] : Manifest](
+    topic: String,
+    brokers: Seq[String]
+  ): KafkaSink[K, V] = {
+    val store = KafkaStore[K, V, KS, VS](topic, brokers)
     lazy val sink = apply[K, V](store)
     sink
```
johnynek: no need for this lazy here... is there? Looks like a misunderstanding of how lazy works.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
johnynek: I looked into this. I found this:
http://stackoverflow.com/questions/11529145/how-do-i-wrap-a-java-util-concurrent-future-in-an-akka-future

so, what we would need to do is something like this:

``` scala
class JavaToScala {
  case class Link[T](future: JFuture[T], p: Promise[T]) {
    def maybeUpdate: Boolean = future.isDone && {
      p.update(Try(future.get))
      true
    }
  }
  def apply[T](jfut: JFuture[T]): Future[T] = {
    val p = new Promise[T]()
    poll(Link(jfut, p))
    p
  }
  private val list = new AtomicReference[List[Link[_]]](Nil)
  private val thread = new Thread(pollRun)
  def start(): Unit = { thread.setDaemon(true); thread.start() }

  private def poll[T](h: Link[T]): Unit = {
    val tail = list.get
    if (list.compareAndSet(tail, h::tail)) ()
    else poll(h)
  }

  private val pollRun = new Runnable {
    def run() = loop(list.getAndSet(Nil))

    @annotation.tailrec
    def loop(items: List[Link[_]]): Unit = {
      val notDone = items.filterNot(_.maybeUpdate)
      if (items.isEmpty || notDone.nonEmpty) Thread.sleep(WAIT)
      loop(list.getAndSet(notDone))
    }
  } 
}
```

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,97 @@
...
+  def apply[K, V](topic: String, props: Properties) = new KafkaStore[K, V](topic, props)
+
+  /**
+    * Create a KafkaStore
+ *
+    * @param topic Kafka topic to produce the messages to
+    * @param brokers Addresses of the Kafka brokers in the hostname:port format
+    * @return Kafka Store
+    */
+  def apply[K, V, KS <: Serializer[K] : Manifest, VS <: Serializer[V] : Manifest](
```
johnynek: why the `Manifest` bound here and not `ClassTag`?

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
johnynek: I think it is worth doing this right. This would mean that you would pass such a `JavaToScala` instance that has a thread inside into the class construction, and you could reuse them across `KafkaStore` instances or have `apply` methods to just allocate a new one (new thread) for each instance (and `release` it in the `close` method).

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaAvroSink.scala
BenFradet: This is legacy code from the current implem.
As stated in the pr desc, the main goal was to upgrade the producer so I haven't really looked into the code, especially for `KafkaSink` and `KafkaAvroSink` and copied them as is.

But since we're here we might as well do a proper refactor.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
BenFradet: Thanks, I'll look into that.

BenFradet: It's too bad there are no bijections available between java futures and twitter futures like there is between scala futures and twitter futures.

johnynek: Indeed. Once we have it here, we should copy it to bijection, but waiting to get a version of bijection is a pain.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,117 @@
...
+  override def put(kv: (K, V)): Future[Unit] = jFutureToTFutureConverter {
+    val (key, value) = kv
+    val f = producer.send(new ProducerRecord[K, V](topic, key, value))
+    new JFuture[Unit] {
+      override def isCancelled: Boolean = f.isCancelled
+      override def get(): Unit = f.get()
+      override def get(timeout: Long, unit: TimeUnit): Unit = f.get(timeout, unit)
+      override def cancel(mayInterruptIfRunning: Boolean): Boolean = f.cancel(mayInterruptIfRunning)
+      override def isDone: Boolean = f.isDone
+    }
```
BenFradet: Here's the "hack" I found to turn a `Future[RecordMetadata]` into a `Future[Unit]`, input welcome.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,117 @@
...
+  
+  /**
+    * Put a key/value pair in a Kafka topic and retrieve record metadata
+    * @param kv (key, value)
+    * @return Future[RecordMetadata]
+    */
+  def putAndRetrieveMetadata(kv: (K, V)): Future[RecordMetadata] = jFutureToTFutureConverter {
+    val (key, value) = kv
+    producer.send(new ProducerRecord[K, V](topic, key, value))
+  }
```
BenFradet, line 59: I think we should also support another version of `put` which gives us back a `Future[RecordMetadata]` like the original kafka api.

BenFradet:
@johnynek I started using your converter (thanks a lot btw), I still have to write a spec for it

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
BenFradet: nvm, I just used a `map`, way better

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+    * Close this store and release any resources.
+    * It is undefined what happens on put/multiGet after close
+    */
+  override def close(time: Time): Future[Unit] = Future {
+    producer.close()
```
johnynek, line 67: we should probably stop the thread converting java to twitter futures (and complete all the open promises with some exception).

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
```diff
@@ -0,0 +1,108 @@
...
+  * Store capable of writing to a Kafka topic.
+  * @author Mansur Ashraf
+  * @since 11/22/13
+  */
+class KafkaStore[K, V](topic: String, props: Properties)
+  extends WritableStore[K, V] with Serializable {
+
+  private lazy val producer = new KafkaProducer[K, V](props)
+  private lazy val jFutureToTFutureConverter = {
+    val converter = new JavaFutureToTwitterFutureConverter
```
johnynek: should we take the timeout as argument? Always using 1 second seems like it could cause issues to me, now that I think about it.

johnynek:
I wonder if the failing test could be due to `close` not making sure all futures complete with something.

BenFradet:
Either that, or the `eventually` logic doesn't work with `poll`, I'm currently investigating.

But I agree, that we should close things properly.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/KafkaStore.scala
BenFradet: yup, agreed

BenFradet:
> Failed tests:
> [error]     com.twitter.storehaus.cache.MutableTTLCacheProperties

+

> Failed tests:
> [error]     com.twitter.storehaus.hbase.HBaseStringStoreProperties

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+  def apply[T](javaFuture: JFuture[T]): Future[T] = {
+    val promise = new Promise[T]()
+    poll(Link(javaFuture, promise))
+    promise
+  }
+  
+  private val pollRun = new Runnable {
+    override def run(): Unit =
+      try {
+        while (!Thread.currentThread().isInterrupted)
```
johnynek: I don't see how `while` is in play here rather than `if`

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+  
+  def stop(): Unit = {
+    list.get().foreach { l =>
+      l.promise.setException(new Exception("Promise not completed"))
+      l.future.cancel(true)
+    }
+    thread.interrupt()
+  }
+
+  private def poll[T](link: Link[T]): Unit = {
```
johnynek: there is a race here if some one calls apply after stop. That could be okay if we advertise that they must be synchronized, or we could handle by having a bigger state:

``` scala
sealed abstract class State
case object Closed extends State
case class Open(links: List[Link[_]]) extends State
val EmptyOpen: State = Open(Nil)

// use this in run, and don't loop is the bool is no true
@annotation.tailrec
def swapOpen(old: List[Link[_]]): (List[Link[_]], Boolean) = list.get match {
  case Closed => (old, false)
  case s@Open(items) = if (list.compareAndSet(s, Open(old))) (items, true) else takeOpen
}

def poll[T](l: Link[T]): Unit = list.get match {
  case Closed => // cancel and complete
  case s@Open(tail) => if (list.compareAndSet(s, Open(l :: tail)) () else poll(l)
}

def stop(): Unit = {
  list.getAndSet(Closed) match {
    case Closed => // already closed
    case Open(ls) => // close all the futures.
  }
}
```

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+    val tail = list.get()
+    if (list.compareAndSet(tail, link :: tail)) ()
+    else poll(link)
+  }
+  
+  case class Link[T](future: JFuture[T], promise: Promise[T]) {
+    def maybeUpdate: Boolean = future.isDone && {
+      promise.update(Try(future.get()))
+      true
+    }
```
johnynek: should add the `cancel` method here.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
BenFradet: Thanks for pointing that out, will look into your solution.

BenFradet: I'm not sure I quite get how the `swapOpen` should fit in `run`.

BenFradet: Here's what I have so far:

``` scala
    override def run(): Unit =
      try {
        if (!Thread.currentThread().isInterrupted)
          loop(list.getAndSet(EmptyState))
      } catch {
        case e: InterruptedException =>
      }

    @tailrec
    def loop(state: State): Unit = state match {
      case s@Open(links) =>
        val notDone = links.filterNot(_.maybeUpdate)
        if (links.isEmpty || notDone.nonEmpty) Thread.sleep(waitTimeMs)
        val (items, open) = swapOpen(notDone)
        if (open) loop(list.getAndSet(Open(items)))
    }
```

BenFradet:
@johnynek I tried to incorporate your comments.
I'm still unsure about how to use `swapOpen` though, input welcome

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -24,60 +24,82 @@ import com.twitter.util.{Future, Try, Promise}
...
+        if (list.compareAndSet(s, Open(old))) (links, true)
+        else swapOpen(links)
+    }
+
+    @tailrec
+    def loop(state: State): Unit = state match {
+      case s@Open(links) =>
+        val notDone = links.filterNot(_.maybeUpdate)
+        if (links.isEmpty || notDone.nonEmpty) Thread.sleep(waitTimeMs)
+        loop(list.getAndSet(Open(notDone)))
```
johnynek: I think this needs to be `swapOpen` and if you get false, you need to `foreach(_.cancel)`, right?

``` scala
swapOpen(notDone) match {
  case (cancels, false) => cancels.foreach(_.cancel)
  case (next, true) => loop(Open(next))
}
```

johnynek:
one minor comment then I am ready to merge (sorry for the delay).

BenFradet:
Thanks for the hint.
However, I'm on holidays this week so I won't be able to take your feedback into account until next week.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,105 @@
...
+    override def run(): Unit =
+      try {
+        if (!Thread.currentThread().isInterrupted)
+          loop(list.getAndSet(EmptyState))
+      } catch {
+        case e: InterruptedException =>
+      }
+
+    @tailrec
+    def swapOpen(old: List[Link[_]]): (List[Link[_]], Boolean) = list.get match {
```
johnynek: actually, I think returning `Option[List[Link[_]]]` would be more idiomatic. When you get `None` you still have the argument. `Boolean` can be confusing what is meant.

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
johnynek: or with `swapOpen(n: List[Link[_]]): Option[List[Link[_]]] = ...`

``` scala
swapOpen(notDone) match {
  case None => notDone.foreach(_.cancel)
  case Some(next) => loop(Open(next))
}
```

storehaus-kafka/src/main/scala/com/twitter/storehaus/kafka/JavaFutureToTwitterFutureConverter.scala
```diff
@@ -0,0 +1,105 @@
...
+    thread.setDaemon(true)
+    thread.start()
+  }
+
+  def stop(): Unit = {
+    list.getAndSet(Closed) match {
+      case Closed => // already closed
+      case s@Open(links) => links.foreach(_.cancel())
+    }
+    thread.interrupt()
```
johnynek: I don't think you need this since once `Closed` shows up, `loop` should terminate.

BenFradet:
[[error]    com.twitter.storehaus.cache.MutableTTLCacheProperties](https://travis-ci.org/twitter/storehaus/jobs/123715431#L944)

@johnynek thanks for the help!

johnynek:
Thanks a ton!

BenFradet:
@johnynek Thanks for merging this.

Do you have a release planned? Because we'd like to use this feature pretty soon.

Also, should we open a pr over at twitter bijection for the future converter?

johnynek:
@BenFradet yeah, it would be great to move the future converter downstream.

We can make a release of storehaus this week, I think. Ping me again if we don't get one out.

BenFradet:
@johnynek great, thanks.

Regarding the pr: will do.

BenFradet:
@johnynek sorry to be a pain but would you mind doing a release?

johnynek:
sorry for the delay.

Pinged the others on the team:
https://gitter.im/twitter/scalding?at=573675c51794136a7d088a0c

if no one objects, I'll publish monday.

BenFradet:
@johnynek No problem and thanks a bunch.

johnynek:
Sorry for the delay, 0.14.0-RC1 is published now, you should be able to see it soon.

BenFradet:
Thanks a lot!

PR #288, opened 2016-03-20
Fix for issue 287

pankajroark:
The root cause is that pattern match on a stream evaluates both head and
tail. So in:
def findT(pred: T => Boolean): Future[T] =
futures match {
  case Stream.Empty => Future.exception(new RuntimeException("Empty iterator in FutureOps.find"))
  case last #:: Stream.Empty => last
  case next #:: rest => next.filter(pred).rescue { case _: Throwable => find(rest)(pred) }
}

when the second case is hit i.e. " case last #:: Stream.Empty => last"
it evaluates the tail, resulting in a call to the second store.

Fixed by eliminating the use of pattern match.

### DIFF
--- a/storehaus-core/src/main/scala/com/twitter/storehaus/FutureOps.scala
+++ b/storehaus-core/src/main/scala/com/twitter/storehaus/FutureOps.scala
@@ -81,12 +81,19 @@ object FutureOps {
     * returns the first Future to both succeed and pass the supplied
     * predicate.
     */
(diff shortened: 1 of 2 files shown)


pankajroark:
Not sure why test failed on travis-ci for scala 2.11.7. The same test passes for me locally with scala 2.11.7.

BenFradet:
Yeah, it's not related to your changes, I have the same problem in #286

johnynek:
Thanks @pankajroark test failure seems totally unrelated, so I'll merge.

PR #289, opened 2016-03-26
Fix failing build

BenFradet:
This PR aims to stop builds from failing on the travis by fixing a test in [MutableCacheProperties](https://github.com/twitter/storehaus/blob/develop/storehaus-cache/src/test/scala/com/twitter/storehaus/cache/MutableTTLCacheProperties.scala).

### DIFF
--- a/storehaus-cache/src/test/scala/com/twitter/storehaus/cache/MutableTTLCacheProperties.scala
+++ b/storehaus-cache/src/test/scala/com/twitter/storehaus/cache/MutableTTLCacheProperties.scala
@@ -37,33 +37,23 @@ object MutableTTLCacheProperties extends Properties("MutableTTLCache") {
 
   def spinSleep(howLong: Long) {
     val finishTime = System.currentTimeMillis + howLong
-    while(System.currentTimeMillis < finishTime){}
+    while (System.currentTimeMillis < finishTime) {}
   }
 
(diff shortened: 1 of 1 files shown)


storehaus-cache/src/test/scala/com/twitter/storehaus/cache/MutableTTLCacheProperties.scala
```diff
@@ -37,33 +37,23 @@ object MutableTTLCacheProperties extends Properties("MutableTTLCache") {
...
+      items.iterator.forall(item => cache.get(item).isEmpty)
+    }
 
-  property("if you put with TTL t, and wait 0 time, it should always (almost, except due to scheduling) be in there") = forAll { (ttl: NegativeTTLTime, items: List[Long])  =>
-      val cache = MutableCache.ttl[Long, Long](ttl.get.milliseconds, items.size + 2)
-      items.foreach{ item =>
-        cache += (item, item)
-      }
-      val sleepTime = ttl.get - 10
-      if(sleepTime > 0) spinSleep(10)
```
BenFradet, line 60: I removed this test since it's always true because the `genNegativeTTLTime` generate ttls between 80 and 480.

BenFradet:
welp, I thought it'd work but apparently not: https://travis-ci.org/twitter/storehaus/jobs/118855150

BenFradet:
I wonder if we should just put the delta directly as the low/upper bound of the generators:
- 40ms for the positive generator
- 80ms for the negative generator

What do you think @johnynek ?

PR #301, opened 2016-06-14
Fix the next version. The convention is to call it SNAPSHOT.

pankajroark:
While doing the release I didn't realize that the convention is to call the next version a SNAPSHOT. Fixing that here.


johnynek:
👍

PR #303, opened 2016-06-15
change KafkaProducer API call

### LINKED ISSUE #302: change KafkaProducer API call
@BenFradet someone pointed out this API to me:

``` java
public java.util.concurrent.Future<RecordMetadata> send(ProducerRecord<K,V> record,
                                               Callback callback)
```

with that API we can convert to a twitter (or scala) future without a timer.

What we do is this:
1. create a `Promise[Unit]`
2. create a `Callback` that has the Promise. When the `onComplete` is called, if the exception in the callback is null, complete the promise normally, otherwise set the exception in the promise.
3. return the `Promise` in 1. as the future.

This work you did was still quite interesting because it resulted in the bijection work, which applies to cases where there is no callback. But here, we should probably just leverage the callback with a Promise.

BenFradet:
fixes #302


johnynek:
amazing! Nice 10x reduction (~200 lines removed to add 20).

Issue #96, opened 2013-04-29
LRUCache and TTLCache should wrap Cache instances, not Map


sritchie:
This method of construction would make it easy to, say, instantiate a `TTLCache` with an `LRUCache`, gaining the semantics of both caches.

PR #305, opened 2016-06-18
wrap Cache instead of Map in LRU, TTL, and LIRS caches

acjay:
Addresses #96. Also lets the current time getter be passed via the convenience constructor for the TTL caches. For testing, I find it helpful to have explicit control over the impure time getter function, rather than having it be internal. It wouldn't be a big deal to just use the regular constructor for `TTLCache`, but for `MutableTTLCache`, the convenience constructor does a lot more.

### DIFF
--- a/storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
+++ b/storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
@@ -30,17 +30,17 @@ import scala.annotation.tailrec
  */
 
 object LIRSCache {
-  def apply[K, V](maxSize:Int, sPercent:Double, backingMap:Map[K, V] = Map.empty[K,V]) = {
+  def apply[K, V](maxSize:Int, sPercent:Double, backingCache:Cache[K, V] = MapCache(Map.empty[K,V])) = {
     val sSize = (maxSize * sPercent).toInt
     val qSize = maxSize - sSize
     require(sSize > 0, "Size of S stack in cache must be >0")
     require(qSize > 0, "Size of Q stack in cache must be >0")
-    new LIRSCache[K, V](LIRSStacks[K](sSize, qSize), backingMap)
+    new LIRSCache[K, V](LIRSStacks[K](sSize, qSize), backingCache)
   }
 }
 
-class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends Cache[K, V] {
-  def get(k: K): Option[V] = backingMap.get(k)
+class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingCache: Cache[K, V]) extends Cache[K, V] {
+  def get(k: K): Option[V] = backingCache.get(k)
 
   def put(kv: (K, V)): (Set[K], Cache[K, V]) = {
     val (k, v) = kv
@@ -50,7 +50,7 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
         // In this case, S is not full, so we just add our key to S, and put the kv in the cache
         val (older, newLirsStacks) = lirsStacks.putOnTopOfS(k)
         older match {
-          case None => (Set.empty[K], new LIRSCache(newLirsStacks, backingMap + kv))
+          case None => (Set.empty[K], new LIRSCache(newLirsStacks, backingCache + kv))
           case _ => throw new IllegalStateException("Stack was not full, yet evicted when element was added")
         }
       } else {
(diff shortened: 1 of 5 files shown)


johnynek:
This looks really great, thanks for tackling this. The only thing I worry about is that we should use `Cache#put` and `Cache#evict` rather that `+` and `-` since they can have the side effect of removing other keys, I think we need to possibly remove those from the data structures in TTL/LIRS, otherwise we may have a memory leak.

storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
```diff
@@ -50,7 +50,7 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
         // In this case, S is not full, so we just add our key to S, and put the kv in the cache
         val (older, newLirsStacks) = lirsStacks.putOnTopOfS(k)
         older match {
-          case None => (Set.empty[K], new LIRSCache(newLirsStacks, backingMap + kv))
+          case None => (Set.empty[K], new LIRSCache(newLirsStacks, backingCache + kv))
```
johnynek, line 53: I think we need `put` here and we need to handle key eviction.

storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
```diff
@@ -61,8 +61,8 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
             // We then add it to S, and if we get something out, take it out of the cache if it is also not in Q. We add kv to the cache.
             val (evictedFromQ, newLirsStacks2) = newLirsStacks1.putOnTopOfQ(k)
             val (evicted1, newBackingMap1) = evictedFromQ match {
-              case Some(oldK) => (Set(oldK), backingMap - oldK)
-              case None => (Set.empty[K], backingMap)
+              case Some(oldK) => (Set(oldK), backingCache - oldK)
```
johnynek, line 64: I think we need `backingCache.evict` here.

storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
```diff
@@ -81,12 +81,12 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
               case Some(oldKinS) => {
                 val (evictedFromQ, newLirsStacks3) = newLirsStacks2.putOnTopOfQ(oldKinS)
                 val (evicted, newBackingMap1) = evictedFromQ match {
-                  case Some(oldKinQ) => (Set(oldKinQ), backingMap - oldKinQ)
-                  case None => (Set.empty[K], backingMap)
+                  case Some(oldKinQ) => (Set(oldKinQ), backingCache - oldKinQ)
```
johnynek, line 84: I think we need `evict` here

storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
```diff
@@ -139,8 +139,8 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
         }
         val (evictedFromS, newLirsStacks2) = newLirsStacks.putOnTopOfS(k)
         val newBackingMap = evictedFromS match {
-          case Some(oldK) if !newLirsStacks2.isInQ(k) => backingMap - oldK
-          case _ => backingMap
+          case Some(oldK) if !newLirsStacks2.isInQ(k) => backingCache - oldK
```
johnynek, line 142: evict.

storehaus-cache/src/main/scala/com/twitter/storehaus/cache/LIRSCache.scala
```diff
@@ -149,11 +149,11 @@ class LIRSCache[K, V](lirsStacks:LIRSStacks[K], backingMap: Map[K, V]) extends C
     }.getOrElse(this)
 
   def evict(k: K): (Option[V], Cache[K, V]) =
-    (get(k), new LIRSCache(lirsStacks.evict(k), backingMap - k))
+    (get(k), new LIRSCache(lirsStacks.evict(k), backingCache - k))
```
johnynek, line 152: backingCache.evict (and clean up the other keys removed from lirsStacks.

acjay:
Seems I missed the response with comments here. Just reminded because of the CLA Assistant. But I no longer work at the company where we used storehaus, so I probably won't get around to finishing this PR. I'm closing, since I imagine it's probably not that in-demand since there are no further comments here or on the original issue.

BrianPin:
Oh man,  I am wish you could merge this to the master, at the bottom line, a TTL + LRU is very useful.
Why not just go ahead to merge it?

PR #307, opened 2016-06-25
Fixed storehaus-caliper compilation

### LINKED ISSUE #281: storehaus-caliper does not compile
Due to a recent refactor, storehaus-caliper fails to compile with the following error:
```
[error] /Users/skrishnan/Projects/storehaus/storehaus-caliper/src/main/scala/com/twitter/storehaus/caliper/WriteThroughCacheBenchmark.scala:65: not found: type HHFilteredStore
[error]     store = (new WriteThroughStore(delayedStore, new HHFilteredStore(new ConcurrentHashMapStore[Long, HLL]), true)).toMergeable
[error]                                                      ^
[error] one error found
```

Looks like we aren't building storehaus-caliper on Travis - hence, we didn't discover this earlier.

BenFradet:
fixes #281

### DIFF
--- a/storehaus-caliper/src/main/scala/com/twitter/storehaus/caliper/WriteThroughCacheBenchmark.scala
+++ b/storehaus-caliper/src/main/scala/com/twitter/storehaus/caliper/WriteThroughCacheBenchmark.scala
@@ -6,10 +6,11 @@ import com.twitter.bijection._
 import com.twitter.storehaus._
 import com.twitter.conversions.time._
 import com.twitter.algebird._
-import java.nio.ByteBuffer
+import com.twitter.storehaus.cache._
+
 import scala.math.pow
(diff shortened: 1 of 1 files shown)


BenFradet:
@johnynek should we add `storehausCaliper` in [build.sbt](https://github.com/twitter/storehaus/blob/develop/build.sbt#L146) so that it gets compiled by travis as well?

BenFradet:
It seems there is a conflict between hbase's zk and kafka's when the two suites are run simultaneously.
I'll investigate.

johnynek:
Great thanks for looking at this!

BenFradet:
From what I've gathered in #308 , we'll need to upgrade hbase in order to force the port (I've been meaning to upgrade storehaus-hbase anyway).
Anyway, should we put storehaus-caliper in the build?

PR #309, opened 2016-07-02
Scalastyle and remove warnings

BenFradet:
This PR introduces scala style and remove a lot of compilation warnings.


johnynek:
This is really good and valuable work, but I am want to ask three things:

1) can we break this up into a project by project change so the diff is reviewable?
2) can we add the mima check as we did here: https://github.com/twitter/algebird/pull/535/files#diff-354f30a63fb0907d4ad57269548329e3R9 to make sure when we are making these changes we don't introduce minor unneeded binary incompatibilities?
3) scalding uses a different linter: https://github.com/twitter/scalding/blob/develop/build.sbt#L113 can we be consistent with that and not use scalastyle? At Twitter, the exact scala style rules caused a lot of debate. Linter seems a little more focused on catching programming errors that getting into style wars.

BenFradet:
1. agreed
2. yup
3. will look into it

PR #312, opened 2016-07-07
Cleanup storehaus-algebra

BenFradet:


storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/Algebras.scala
```diff
@@ -16,29 +16,33 @@
...
+  implicit def semigroup[K, V: Semigroup]: Semigroup[ReadableStore[K, V]] =
+    new ReadableStoreSemigroup[K, V]
   implicit def monoid[K, V: Monoid]: Monoid[ReadableStore[K, V]] = new ReadableStoreMonoid[K, V]
 }
 
 /** A semigroup that uses the future semigroup to add results of both gets */
 class ReadableStoreSemigroup[K, V: Semigroup] extends Semigroup[ReadableStore[K, V]] {
-  override def plus(l: ReadableStore[K, V], r: ReadableStore[K,V]) =
-    new AbstractReadableStore[K,V] {
+  override def plus(l: ReadableStore[K, V], r: ReadableStore[K, V]): AbstractReadableStore[K, V] =
```
johnynek: if we make this return `ReadableStore[K, V]` does it break mima? We don't really want to return that subclass.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/BufferingStore.scala
```diff
@@ -57,13 +59,14 @@ class BufferingMergeable[K, V](store: Mergeable[K, V], summerCons: SummerConstru
         k -> foptV
       }
 
-  override def multiMerge[K1 <: K](kvs: Map[K1, V]) = {
-    val links: Map[K, PromiseLink[V]] = kvs.map { case (k1, v) => k1 -> PromiseLink(v) }(breakOut) // no lazy
-    summer.put(links).foreach(mergeFlush(_))
+  override def multiMerge[K1 <: K](kvs: Map[K1, V]): Map[K1, Promise[Option[V]]] = {
```
johnynek: this should be `Map[K1, Future[Option[V]]` we should definitely not expose the `Promise` which is actually mutable.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/ConcurrentHashMapMergeableStore.scala
```diff
@@ -24,11 +39,11 @@ class ConcurrentHashMapMergeableStore[K, V](map: JConcurrentHashMap[K, V])(
 
   override def merge(kv: (K, V)): Future[Option[V]] = {
     val (k, v) = kv
-    map.get(k) match {
-      case null =>
-        if (null == map.putIfAbsent(k, v)) Future.None
+    Option(map.get(k)) match {
```
johnynek: let's not do this. This is allocating where in the past we were avoiding allocations. This could slow down operations in a path that may be important for performance.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/ConvertedStore.scala
```diff
@@ -20,28 +20,32 @@ import com.twitter.bijection.Injection
...
+ * TODO: we should add a specific exception type here so we can safely filter these cases to
+ * Future.None if we so choose.
  */
 @deprecated("Use com.twitter.storehaus.ConvertedStore instead", "0.3.1")
-class ConvertedStore[K1, -K2, V1, V2](store: Store[K1, V1])(kfn: K2 => K1)(implicit inj: Injection[V2, V1])
+class ConvertedStore[K1, -K2, V1, V2](store: Store[K1, V1])(kfn: K2 => K1)
+  (implicit inj: Injection[V2, V1])
   extends ConvertedReadableStore[K1, K2, V1, V2](store)(kfn)({ v1: V1 =>
-    inj.invert(v1).map { Future.value(_) }
+    inj.invert(v1).map { Future.value }
```
johnynek: can we use `.map(Future.value)` style instead?

storehaus-algebra/src/test/scala/com/twitter/storehaus/algebra/ReadableStoreProperties.scala
```diff
@@ -16,17 +16,17 @@
...
-import com.twitter.util.{ Await, Future }
+import com.twitter.util.Await
 import com.twitter.storehaus.ReadableStore
-import org.scalacheck.{ Arbitrary, Properties }
-import org.scalacheck.Gen.choose
+import org.scalacheck.{Prop, Arbitrary, Properties}
 import org.scalacheck.Prop._
 
-object ReadableStoreAlgebraProperties extends Properties("ReadableStoreAlgebra") {
+object ReadableStoreProperties extends Properties("ReadableStoreAlgebra") {
```
johnynek: why the name change here?

johnynek:
nice work. Couple of small points

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/ConcurrentHashMapMergeableStore.scala
BenFradet: Good to know, I'll fix it.

storehaus-algebra/src/test/scala/com/twitter/storehaus/algebra/ReadableStoreProperties.scala
BenFradet: The file and the class didnt have the same name so I picked one. I can definitely go the other way around though.

BenFradet:
Thanks for your review, I'll address those comments later today.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/Algebras.scala
BenFradet: it does break mima, I'll add an exclusion

johnynek:
Nice work!

PR #318, opened 2016-07-25
Cleanup storehaus-hbase

BenFradet:


storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseByteArrayStore.scala
```diff
@@ -0,0 +1,72 @@
...
+  protected val table: String,
+  protected val columnFamily: String,
+  protected val column: String,
+  protected val client: HBaseClient,
+  protected val threads: Int) extends Store[Array[Byte], Array[Byte]] with AsyncHBaseStore {
+
+  /** get a single key from the store.
+    * Prefer multiGet if you are getting more than one key at a time
+    */
+  override def get(k: Array[Byte]): Future[Option[Array[Byte]]] = {
```
johnynek: don't need `{ }` here or on `put`.

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseByteArrayStore.scala
```diff
@@ -0,0 +1,68 @@
...
+   * replace a value
+   * Delete is the same as put((k,None))
+   */
+  override def put(kv: (Array[Byte], Option[Array[Byte]])): Future[Unit] = putValue(kv)
+
+  /** Close this store and release any resources.
+    * It is undefined what happens on get/multiGet after close
+    */
+  override def close(t: Time): Future[Unit] = futurePool {
+    client.shutdown().join()
```
johnynek: okay, geekout time. Can we not use the `futurePool` here and instead write a function to convert `Deferred[T]` to `Future[T]`?

http://tsunanet.net/~tsuna/async/api/com/stumbleupon/async/Deferred.html?is-external=true

I think we just want something like:

``` scala
def toFuture[T](d: Deferred[T]): Future[T] = {
  val p = new Promise[T]()
  d.addCallbacks(new Callback[Unit, T] {
    def call(t: T) = p.updateIfEmpty(Return(t))
  }, new Callback[Unit, Exception] {
    // looks like this has to be an exception
    // https://github.com/OpenTSDB/async/blob/master/src/Deferred.java#L685
    def call(e: Exception) = p.updateIfEmpty(Throw(e))
  })
  // return the promise as a Future
  p
}
```

With this we can just do: `toFuture(client.shutdown()).unit`

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseLongStore.scala
```diff
@@ -63,6 +64,6 @@ class AsyncHBaseLongStore(protected val quorumNames: Seq[String],
   /** Close this store and release any resources.
     * It is undefined what happens on get/multiGet after close
     */
-  override def close(t: Time) = futurePool { client.shutdown() }
+  override def close(t: Time): Future[Unit] = futurePool { client.shutdown() }
```
johnynek: same comment here. In fact, I think this is more wrong since I think this is not actually waiting till the client has closed.

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseStringStore.scala
```diff
@@ -60,7 +62,7 @@ class AsyncHBaseStringStore(protected val quorumNames: Seq[String],
   /** Close this store and release any resources.
     * It is undefined what happens on get/multiGet after close
     */
-  override def close(t: Time) = futurePool {
+  override def close(t: Time): Future[Unit] = futurePool {
```
johnynek: same comment.

storehaus-hbase/src/main/scala/com/twitter/storehaus/hbase/HBaseByteArrayStore.scala
```diff
@@ -74,5 +82,5 @@ class HBaseByteArrayStore(protected val quorumNames: Seq[String],
   /** Close this store and release any resources.
     * It is undefined what happens on get/multiGet after close
     */
-  override def close(t: Time) = futurePool { pool.close() }
+  override def close(t: Time): Future[Unit] = futurePool { pool.close() }
```
johnynek, line 85: same comment.

johnynek, line 85: actually this is different.

storehaus-hbase/src/main/scala/com/twitter/storehaus/hbase/HBaseStore.scala
```diff
@@ -41,7 +41,7 @@ trait HBaseStore {
   protected val futurePool = FuturePool(Executors.newFixedThreadPool(threads))
```
johnynek, line 41: we should have a todo to pass in the `FuturePool`. Should not make these internally

storehaus-hbase/src/main/scala/com/twitter/storehaus/hbase/HBaseStore.scala
```diff
@@ -41,7 +41,7 @@ trait HBaseStore {
...
 
   def getHBaseAdmin: HBaseAdmin = {
-    if (conf.get("hbase.zookeeper.quorum") == null) {
+    if (Option(conf.get("hbase.zookeeper.quorum")).isEmpty) {
```
johnynek: if you are going this route, use `.foreach`:

``` scala
Option(conf.get("hbase.zookeeper.quorum")).foreach { _ =>
  conf.set("hbase.zookeeper.quorum", quorumNames.mkString(","))
}
```

storehaus-hbase/src/main/scala/com/twitter/storehaus/hbase/HBaseStore.scala
```diff
@@ -65,7 +65,10 @@ trait HBaseStore {
     require(isNotEmpty(column), "column is required")
   }
 
-  def getValue[K, V](key: K)(implicit keyInj: Injection[K, Array[Byte]], valueInj: Injection[V, Array[Byte]]): Future[Option[V]] = {
+  def getValue[K, V](key: K)(
+    implicit keyInj: Injection[K, Array[Byte]],
+    valueInj: Injection[V, Array[Byte]]
+  ): Future[Option[V]] = {
     val tbl = pool.getTable(table)
```
johnynek, line 72: why is this outside of the Future, I wonder, and nothing else is?

johnynek, line 72: ahh, because we close it at the end.... got it.

BenFradet:
@johnynek thanks for your review, will update tomorrow.

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseStore.scala
```diff
@@ -20,7 +20,6 @@ import com.twitter.util.{Future, FuturePool}
 import java.util.concurrent.Executors
 import org.hbase.async.{DeleteRequest, PutRequest, GetRequest, HBaseClient}
 import com.twitter.bijection.{Injection, Codec, Conversion}
-import scala.Some
 import Conversion._
 import scala.collection.JavaConverters._
```
johnynek, line 25: actually, below I think we want to use `toFuture[T](d: Deferred[T]): Future[T]` and actually not use the `FuturePool` at all, I think. In fact, I think we could possibly remove the `futurePool` and the `threads` val.

johnynek:
Looks like more issues here we missed in review the first time. If we are going to clean it up, I think we should really address these issues.

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseStore.scala
johnynek, line 25: I wonder if what we really want is a

``` scala
case class Table(asBytes: Array[Byte])
case class Column(asBytes: Array[Byte])
case class ColumnFamily(asBytes: Array[Byte])
case class Key(asBytes: Array[Byte])
case class Result(timestamp: Long, valueAsBytes: Array[Byte])
class AsyncHBaseStore extends Store[(Table, ColumnFamily, Column, Key), List[Result]]
```

and then if you just want the change the key you can do:

``` scala
val baseKey = (tab, cf, col, Key(Array[Byte]()))
val store: AsyncHBaseStore = ...
def get(k: Array[Byte]) = store.get(baseKey.copy(_4 = Key(k)))`
```

or something like this. Not sure why we should keep the table, column, etc... out of the key if people want to use that?

johnynek, line 25: maybe not reuse the same name there, maybe call the above proposal something different.

storehaus-hbase/src/main/scala/com/twitter/storehaus/asynchbase/AsyncHBaseStore.scala
```diff
@@ -36,46 +36,56 @@ trait AsyncHBaseStore {
...
   def validateConfiguration() {
     import org.apache.commons.lang.StringUtils.isNotEmpty
 
-    require(!quorumNames.isEmpty, "Zookeeper quorums are required")
+    require(quorumNames.nonEmpty, "Zookeeper quorums are required")
     require(isNotEmpty(columnFamily), "column family is required")
     require(isNotEmpty(column), "column is required")
     require(isNotEmpty(table), "table is required")
-    client.ensureTableExists(table.as[Array[Byte]])
+    client.ensureTableExists(table)
```
BenFradet, line 46: I tried awaiting the result of this async call, however it hangs forever. As is, it doesn't do anything anyway.

johnynek:
This looks fine to me.

I think you can fix the binary incompatibility by:
1. Making `toFuture` a method on an object that you call.
2. make the deleted futurePool a lazy val and mark it @deprecated

BenFradet:
Regarding `futurePool` shouldn't we introduce exceptions since it's dead code?

johnynek:
@BenFradet well, you're probably right it is bad idea on the futurePool. Let's just add exclusions for Mima for that (and the next version will bump the version number). It is unlikely to hurt anyone unless they subclassed these stores.

BenFradet:
I want to keep this open for a while more, I'm trying to introduce a spec for the async store.

BenFradet:
Also, I'd like to write a few unit tests for `toFuture`

build.sbt
```diff
@@ -50,7 +50,12 @@ val ignoredABIProblems = {
...
     exclude[MissingMethodProblem]("com.twitter.storehaus.kafka.KafkaStore.apply"),
     exclude[IncompatibleResultTypeProblem]("com.twitter.storehaus.algebra.ReadableStoreSemigroup" +
-      ".plus")
+      ".plus"),
+    exclude[MissingMethodProblem]("com.twitter.storehaus.asynchbase.AsyncHBaseByteArrayStore" +
+      ".futurePool"),
+    exclude[MissingMethodProblem]("com.twitter.storehaus.asynchbase.AsyncHBaseLongStore" +
+      ".futurePool"),
+    exclude[MissingMethodProblem]("com.twitter.storehaus.asynchbase.AsyncHBaseStore.futurePool")
   )
```
johnynek, line 62: looks like you missed:

```
[error]  * method futurePool()com.twitter.util.ExecutorServiceFuturePool in class com.twitter.storehaus.asynchbase.AsyncHBaseStringStore does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("com.twitter.storehaus.asynchbase.AsyncHBaseStringStore.futurePool")
```

BenFradet, line 62: Indeed, my bad. Will fix later today.

BenFradet:
Welp, I'll need to dig a bit depper into asynchbase since nothing seems to be working :laughing:

johnynek:
@BenFradet why do you say it is not working? The failures are the crappy twttr maven server (why oh why didn't they just publish to maven central.... actually, I know, but still....)

I just kicked the 2.10 build. Let's see if it works.

BenFradet:
I meant the spec for `AsyncHBaseStore` I'm trying to write.

BenFradet:
@johnynek Could you merge this? I'll add a spec for the asynchbase stores when I'll upgrade hbase.

johnynek:
Yeah, let's move forward. It would be great to improve test coverage, but this seems like an advance.

PR #324, opened 2016-08-07
Cleanup storehaus-mongodb

BenFradet:


BenFradet:
ping @johnynek in case you missed it

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -93,24 +96,24 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
-        if (valueObject.get(valueName) == None) None
-        else throw new ClassCastException("Cannot convert %s to %s".format(valueObject.get(valueName).getClass(), manifest[V]))
-      }
-      case Some(value) => {
-        if (getManifest().isInstance(value)) Some(value)
-        else throw new ClassCastException("Cannot convert %s to %s".format(value.getClass(), manifest[V]))
-      }
+      case None =>
+        if (valueObject.get(valueName).isEmpty) None
+        else throw new ClassCastException("Cannot convert %s to %s"
```
johnynek: can we use `s"Cannot convert ${value.getClass} to ${manifest[V]}"`

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -93,24 +96,24 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
+          .format(valueObject.get(valueName).getClass, manifest[V]))
+      case Some(value) =>
+        if (getManifest.isInstance(value)) Some(value)
+        else throw new ClassCastException("Cannot convert %s to %s"
```
johnynek: same comment.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -93,24 +96,24 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
 
   protected def getValue(valueObject: MongoDBObject): Option[V] = {
```
johnynek: seems like this method should really be returning `Try[V]` rather than an option and throwing. :/

johnynek:
one minor comment (about string formatting).

I would say that maybe this code needs to be reworked. It might make sense to move it to reactivemongo http://reactivemongo.org/ which is already async and handles most of the issues here (but that is out of scope for the current PR).

BenFradet:
Agreed, will tackle string formatting here and create an issue to move to reactive mongo.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
BenFradet: Sorry I haven't been able to work on that for a while, I've been a bit overwhelmed with other things.

I wanted your input regarding this:

``` scala

  protected def getValue(valueObject: MongoDBObject): Try[V] =
    valueObject.getAs[V](valueName) match {
      case None =>
        if (valueObject.get(valueName).isEmpty) // here
        else Throw(
          new ClassCastException(s"Cannot convert ${valueObject.get(valueName)} to ${manifest[V]}"))
      case Some(value) =>
        if (getManifest.isInstance(value)) Return(value)
        else Throw(new ClassCastException(s"Cannot convert ${value.getClass} to ${manifest[V]}"))
    }
```

I see only two options, either we return a `Return` but that will falsify a `toOption` call on the result or we return a `Throw` which shouldn't be thrown, what do you think?

johnynek: looks good to me, I'd just add:

``` scala
s"Cannot convert $value of type ${value.getClass} to ${manifest[V]}"
```

so we can see the full value and type in the exception message.

BenFradet: Ok, but what about the

``` scala
case None =>
  if (valueObject.get(valueName).isEmpty)
```

case?

johnynek: I think it `.isEmpty` is fine if we are never looking at the `Some(contents)` to get the data out.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -86,31 +89,30 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
   override def get(key: K): Future[Option[V]] = futurePool {
     col.findOne(MongoDBObject(keyName -> key)).flatMap { valueObject => {
-        getValue(valueObject)
+        getValue(valueObject).toOption
       }
     }
   }
 
-  protected def getValue(valueObject: MongoDBObject): Option[V] = {
+  protected def getValue(valueObject: MongoDBObject): Try[V] =
```
johnynek: I think we want `Try[Option[V]]` here actually. Some things are legitimately `None`.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -86,31 +89,30 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
-      case None => {
-        if (valueObject.get(valueName) == None) None
-        else throw new ClassCastException("Cannot convert %s to %s".format(valueObject.get(valueName).getClass(), manifest[V]))
-      }
-      case Some(value) => {
-        if (getManifest().isInstance(value)) Some(value)
-        else throw new ClassCastException("Cannot convert %s to %s".format(value.getClass(), manifest[V]))
-      }
+      case None =>
+        if (valueObject.get(valueName).isEmpty) Throw(new Exception("Value not found"))
```
johnynek: I think this should be `Return(None)` on this branch, which is to say it was really empty.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -86,31 +89,30 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
+        else Throw(
+          new ClassCastException(s"Cannot convert ${valueObject.get(valueName)} to ${manifest[V]}"))
+      case Some(value) =>
+        if (getManifest.isInstance(value)) Return(value)
```
johnynek: this should be `Return(Some(value))`

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -86,31 +89,30 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
...
+        else Throw(new ClassCastException(
+          s"Cannot convert $value of type ${value.getClass} to ${manifest[V]}"))
     }
-  }
 
-  protected def getManifest() = {
+  protected def getManifest = {
```
johnynek: let's put the return type `: Class[_]` I think.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
```diff
@@ -86,31 +89,30 @@ class MongoStore[K: MongoValue, V: MongoValue: Manifest] (
 
   override def get(key: K): Future[Option[V]] = futurePool {
     col.findOne(MongoDBObject(keyName -> key)).flatMap { valueObject => {
-        getValue(valueObject)
+        getValue(valueObject).toOption
```
johnynek: this should be:

``` scala
futurePool {
  col.findOne(MongoDBObject(keyName -> key)) match {
    case None => Future.None
    case Some(v) => Future.const(getValue(v))
  }
}.flatten
```

I think.

storehaus-mongodb/src/main/scala/com/twitter/storehaus/mongodb/MongoStore.scala
BenFradet: Yeah, that's what I wanted your input on earlier.

johnynek: Sorry I overlooked that before.

BenFradet: No problem.

BenFradet:
Thanks a lot for your review, sorry if I wasn't clear enough earlier.

johnynek:
👍

PR #327, opened 2016-08-27
Cleanup storehaus-mysql

BenFradet:


BenFradet:
~~@johnynek could you trigger a new build when you get the chance, twitter's mvn is having issues again.~~
Nevermind the failing build was the one before the rebase.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MergeableMySqlStore.scala
```diff
@@ -44,31 +43,31 @@ class MergeableMySqlStore[V](underlying: MySqlStore)(implicit inj: Injection[V,
...
 
         // handle inserts for new keys
-        val insertF = newKeys.isEmpty match {
-          case true => Future.Unit
-          case false =>
+        val insertF =
+          if (newKeys.isEmpty) {
+            Future.Unit
+          } else {
             val insertKvs = newKeys.map { k => k -> kvs.get(k).get }
```
johnynek: can we change this to `k -> kvs(k)` rather than `.get(k).get`?

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MergeableMySqlStore.scala
```diff
@@ -44,31 +43,31 @@ class MergeableMySqlStore[V](underlying: MySqlStore)(implicit inj: Injection[V,
...
-            insertKvs.isEmpty match {
-              case true => Future.Unit
-              case false => underlying.executeMultiInsert(insertKvs.toMap.mapValues { v => inj(v) })
-            }
-        }
+            if (insertKvs.isEmpty) Future.Unit
+            else underlying.executeMultiInsert(insertKvs.toMap.mapValues(v => inj(v))).unit
```
johnynek: what about `insertKvs.toMap.mapValues(inj)`? I think that will work.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MergeableMySqlStore.scala
```diff
@@ -44,31 +43,31 @@ class MergeableMySqlStore[V](underlying: MySqlStore)(implicit inj: Injection[V,
...
         // handle update/merge for existing keys
         // lazy val realized inside of insertF.flatMap
-        lazy val updateF = existingKeys.isEmpty match {
-          case true => Future.Unit
-          case false =>
+        lazy val updateF =
+          if (existingKeys.isEmpty) {
+            Future.Unit
+          } else {
             val existingKvs = existingKeys.map { k => k -> kvs.get(k).get }
```
johnynek: same `k -> kvs(k)` comment

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MergeableMySqlStore.scala
```diff
@@ -44,31 +43,31 @@ class MergeableMySqlStore[V](underlying: MySqlStore)(implicit inj: Injection[V,
...
             underlying.executeMultiUpdate(existingKvs.map { kv =>
               val resV = semigroup.plus(result.get(kv._1).get.get, kv._2)
```
johnynek: looks like a lot of `get` there. what about:

``` scala
existingKvs.map { case (k, v) =>
   k -> inj(semigroup.plus(result(k).get v))
}
```

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
   protected [mysql] def rollbackTransaction : Future[Unit] = client.query(ROLLBACK_TXN_SQL).unit
 
-  protected [mysql] def executeMultiInsert[K1 <: MySqlValue](kvs: Map[K1, MySqlValue]) = {
-    val insertSql = MULTI_INSERT_SQL_PREFIX + Stream.continually("(?, ?)").take(kvs.size).mkString(",")
-    val insertParams = kvs.map { kv =>
-      List(MySqlStringInjection(kv._1).getBytes, MySqlStringInjection(kv._2).getBytes)
+  protected [mysql] def executeMultiInsert[K1 <: MySqlValue](
+      kvs: Map[K1, MySqlValue]): Future[Result] = {
+    val insertSql = MULTI_INSERT_SQL_PREFIX +
+      Stream.continually("(?, ?)").take(kvs.size).mkString(",")
```
johnynek: can we say `Iterator.continually` which I think is slightly faster than a Stream if we are just calling `mkString`?

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
 
-  protected [mysql] def executeMultiUpdate[K1 <: MySqlValue](kvs: Map[K1, MySqlValue]) = {
-    val updateSql = MULTI_UPDATE_SQL_PREFIX + Stream.continually("WHEN ? THEN ?").take(kvs.size).mkString(" ") +
+  protected [mysql] def executeMultiUpdate[K1 <: MySqlValue](
+      kvs: Map[K1, MySqlValue]): Future[Result] = {
+    val updateSql = MULTI_UPDATE_SQL_PREFIX +
+      Stream.continually("WHEN ? THEN ?").take(kvs.size).mkString(" ") +
       MULTI_UPDATE_SQL_INFIX + Stream.continually("?").take(kvs.size).mkString("(", ",", ")")
 
     val updateParams = kvs.map { kv =>
```
johnynek: can we use `case (k, v) =>` here?

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
-      (MySqlStringInjection(kv._1).getBytes, MySqlStringInjection(kv._2).getBytes)
+      (String2MySqlValueInjection.invert(kv._1).getOrElse("").getBytes,
+        String2MySqlValueInjection.invert(kv._2).getOrElse("").getBytes)
     }
     // params for "WHEN ? THEN ?"
-      val updateCaseParams: Seq[Parameter] =
-        updateParams.map { kv => List(kv._1, kv._2) }.toSeq.flatten.map(Parameter.wrap[Array[Byte]])
+    val updateCaseParams: Seq[Parameter] =
+      updateParams.map { kv => List(kv._1, kv._2) }.toSeq.flatten
```
johnynek: `flatMap { case (k, v) => List(k, v) }` I think we will work if you keep the return type annotation.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
-    val updateSql = MULTI_UPDATE_SQL_PREFIX + Stream.continually("WHEN ? THEN ?").take(kvs.size).mkString(" ") +
+  protected [mysql] def executeMultiUpdate[K1 <: MySqlValue](
+      kvs: Map[K1, MySqlValue]): Future[Result] = {
+    val updateSql = MULTI_UPDATE_SQL_PREFIX +
+      Stream.continually("WHEN ? THEN ?").take(kvs.size).mkString(" ") +
       MULTI_UPDATE_SQL_INFIX + Stream.continually("?").take(kvs.size).mkString("(", ",", ")")
 
     val updateParams = kvs.map { kv =>
-      (MySqlStringInjection(kv._1).getBytes, MySqlStringInjection(kv._2).getBytes)
+      (String2MySqlValueInjection.invert(kv._1).getOrElse("").getBytes,
```
johnynek: pretty skeptical about the `.getOrElse("")` can you think why that is safe? If we can't invert, shouldn't we be passing that up or something? We are in a `Future` so we can give a failed result. I think we need to lift the `Try` into a `Future` at this point and flatMap.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
     // to convert resultset into desired output format
     // we assume here the mysql client already has the dbname/schema selected
     val mysqlResult =
-      selectStmt.select(MySqlStringInjection(k).getBytes) { row =>
+      selectStmt.select(String2MySqlValueInjection.invert(k).getOrElse("").getBytes) { row =>
         row(vCol).map { MySqlValue(_)}
       }
 
-    mysqlResult.map { result => result.lift(0).flatten.headOption }
+    mysqlResult.map(result => result.headOption.flatten)
```
johnynek: `mysqlResult(_.headOption.flatten)`

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+    client.prepare(updateSql)(updateCaseParams ++ updateInParams: _*)
   }
 
   override def get(k: MySqlValue): Future[Option[MySqlValue]] = {
     // finagle-mysql select() method lets you pass in a mapping function
     // to convert resultset into desired output format
     // we assume here the mysql client already has the dbname/schema selected
     val mysqlResult =
-      selectStmt.select(MySqlStringInjection(k).getBytes) { row =>
+      selectStmt.select(String2MySqlValueInjection.invert(k).getOrElse("").getBytes) { row =>
```
johnynek: let't not use `getOrElse`, let's properly lift that into the `Future`.

``` scala
// put this in some package object if we don't already have it somewhere.
def toTwitterFuture[T](t: Try[T]): Future[T] = t match {
  case Success(t) => Future.value(t)
  case Failure(e) => Future.exception(e)
}

toFuture(String2MySqlValueInjection.invert(k)).flatMap { str =>
  selectStmt.select(str.getBytes) { row => row(vCol).map(MySqlValue(_)) }
}

```

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
-      mysqlResult.map { rows =>
-        rows.toMap.filterKeys { _ != None }.map { case (optK, optV) => (optK.get, optV) }
-      },
-      { (k: K1) => Future.None }
-    )
+    if (ks.isEmpty) {
+      Map()
+    } else {
+      // build preparedstatement based on keyset size
+      val placeholders = Stream.continually("?").take(ks.size).mkString("(", ",", ")")
```
johnynek: `Iterator.continually`

johnynek: Can we share this utility function with the other one?

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+      val selectSql = MULTI_SELECT_SQL_PREFIX + placeholders
+
+      val params = ks.map(String2MySqlValueInjection.invert(_).getOrElse("").getBytes).toSeq
```
johnynek: no getOrElse here.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -97,59 +100,68 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+        .map(Parameter.wrap[Array[Byte]])
+      val mysqlResult =
+        client.prepare(selectSql).select(params: _*) { row =>
+          (row(kCol).map(MySqlValue(_)), row(vCol).map(MySqlValue(_)))
+        }
+      FutureOps.liftValues(
+        ks,
+        mysqlResult.map { rows =>
+          rows.toMap.filterKeys(_.isDefined).map { case (optK, optV) => (optK.get, optV) }
```
johnynek:
``` scala
rows.iterator.collect { case (Some(k), v) => (k, v) }.toMap
```

is better

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -161,62 +173,70 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+          } else {
             // do not include None values in insert query
-            val insertKvs = newKeys.map { k => k -> kvs.getOrElse(k, None) }.filter { ! _._2.isEmpty }
-              .toMap.mapValues { v => v.get }
-            insertKvs.isEmpty match {
-              case true => Future.Unit
-              case false => executeMultiInsert(insertKvs)
-            }
-        }
+            val insertKvs = newKeys.map(k => k -> kvs.getOrElse(k, None))
```
johnynek:
``` scala
newKeys.iterator.collect {case k if kvs.contains(k) => k -> kvs(k) }.toMap
```

I think is better (one pass over the data, not 3)

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -161,62 +173,70 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+        val existingKvs = existingKeys.map(k => k -> kvs.getOrElse(k, None))
 
         // do not include None values in update query
-        val updateKvs = existingKvs.filter { ! _._2.isEmpty }
-          .toMap.mapValues { v => v.get }
-        lazy val updateF = updateKvs.isEmpty match {
-          case true => Future.Unit
-          case false => executeMultiUpdate(updateKvs)
-        }
+        val updateKvs = existingKvs.filter(_._2.isDefined)
```
johnynek:
``` scala
existingKvs.iterator.collect { case (k, Some(v)) => (k, v) }.toMap
```

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -161,62 +173,70 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
         // deletes
-        val deleteKeys = existingKvs.filter { _._2.isEmpty }.map { _._1 }
-        lazy val deleteF = deleteKeys.isEmpty match {
-          case true => Future.Unit
-          case false =>
-            val deleteSql = MULTI_DELETE_SQL_PREFIX + Stream.continually("?").take(deleteKeys.size).mkString("(", ",", ")")
-            val deleteParams = deleteKeys.map { k => MySqlStringInjection(k).getBytes }.toSeq.map(Parameter.wrap[Array[Byte]])
-            client.prepare(deleteSql)(deleteParams:_*)
-        }
+        val deleteKeys = existingKvs.filter(_._2.isEmpty).map(_._1)
```
johnynek:
``` scala
existingKvs.iterator.collect { case (k, None) => k }.toVector
```

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -161,62 +173,70 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+              }.toSeq
+            client.prepare(deleteSql)(deleteParams: _*).unit
+          }
 
         // sequence the three queries. the inner futures are lazy
-        insertF.flatMap { f =>
-          updateF.flatMap { f =>
-            deleteF.flatMap { f => commitTransaction }
-              .handle { case e: Exception => rollbackTransaction.flatMap { throw e } }
+        insertF.flatMap { _ =>
```
johnynek: can we do:

``` scala
def commitOrRollback(): Future[Unit] =
  commitTransaction
    .handle { case e: Throwable => rollbackTransaction.flatMap { _ => Future.exception(e) } }

for {
_ <- insertF
_ <- updateF
_ <- deleteF
_ <- commitOrRollback
} yield ()
```

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -226,20 +246,19 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
-    get(k).flatMap { optionV =>
-      optionV match {
-        case Some(value) =>
-          updateStmt(MySqlStringInjection(v).getBytes, MySqlStringInjection(k).getBytes)
-        case None =>
-          insertStmt(MySqlStringInjection(k).getBytes, MySqlStringInjection(v).getBytes)
-      }
+    get(k).flatMap {
+      case Some(value) =>
+        updateStmt(String2MySqlValueInjection.invert(v).getOrElse("").getBytes,
```
johnynek: let's fix the `getOrElse`

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -226,20 +246,19 @@ class MySqlStore(protected [mysql] val client: Client, table: String, kCol: Stri
...
+        insertStmt(String2MySqlValueInjection.invert(k).getOrElse("").getBytes,
+          String2MySqlValueInjection.invert(v).getOrElse("").getBytes)
     }
   }
 
-  protected def doDelete(k: MySqlValue): Future[Result] = {
-    deleteStmt(MySqlStringInjection(k).getBytes)
-  }
+  protected def doDelete(k: MySqlValue): Future[Result] =
+    deleteStmt(String2MySqlValueInjection.invert(k).getOrElse("").getBytes)
```
johnynek: getOrElse has to go

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/ValueMapper.scala
```diff
@@ -78,7 +78,8 @@ object ValueMapper {
       case EmptyValue => Some("")
       case NullValue => None
       // all other types are currently unsupported
-      case _ => throw new UnsupportedOperationException(v.getClass.getName + " is currently not supported.")
+      case _ => throw new UnsupportedOperationException(value.getClass.getName +
```
johnynek: can we do `case other =>` and include `$other` in the error message? Helps debugging usually.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/ValueMapper.scala
```diff
@@ -101,7 +103,8 @@ object MySqlValue {
       v.readBytes(bytes)
       v.resetReaderIndex()
       new MySqlValue(RawValue(Type.Blob, Charset.Binary, isBinary = true, bytes))
-    case _ => throw new UnsupportedOperationException(v.getClass.getName + " is currently not supported.")
+    case _ => throw new UnsupportedOperationException(v.getClass.getName +
```
johnynek: lets include the value in the error message.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
BenFradet: I didn't know about that, if you don't mind would you care to explain why?

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
BenFradet: To be honest, I just wanted to get rid of [the deprecation warning](https://github.com/twitter/storehaus/blob/develop/storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/ValueMapper.scala#L134) but I guess the `getOrElse` was the reason it was deprecated in the first place :laughing:

BenFradet:
Thanks for your review, I think I've taken care of everything you mentioned.

I'd still like to know why prefer `Iterator.continually` over `Stream.continually`.

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/package.scala
```diff
@@ -0,0 +1,35 @@
...
+    case Success(v) => Future.value(v)
+    case Failure(ex) => Future.exception(ex)
+  }
+  def toTwitterFuture[T](ts: Seq[Try[T]]): Future[Seq[T]] =
+    if (ts.forall(_.isSuccess)) Future.value(ts.map(_.get))
+    else ts.find(_.isFailure).head match {
+      case Failure(ex) => Future.exception(ex)
+      case _ =>
+        Future.exception(new Exception(s"Unknown error when converting $ts to future of seq"))
+    }
```
BenFradet: @johnynek I'd like to know what you think about this.

johnynek:
This may be superstition but when I know I will consume the collection, I
convert to Iterator and do the operation on that. The memorization of
Stream I assume to be not free.

What said, I should benchmark it.

PS: I'm on vacation this week so maybe someone else can merge or I can get
to it when I get back.

BenFradet:
Ok no worries I'll be waiting till next week.
Have a great holiday!

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/package.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.twitter.util.Future
+
+import scala.util.{Failure, Success, Try}
+
+package object mysql {
+  def toTwitterFuture[T](t: Try[T]): Future[T] = t match {
+    case Success(v) => Future.value(v)
+    case Failure(ex) => Future.exception(ex)
+  }
+  def toTwitterFuture[T](ts: Seq[Try[T]]): Future[Seq[T]] =
```
johnynek: I'd rather:

``` scala
val bldr = Vector.newBuilder[T]
val tryRes = ts.iterator.foldLeft(Try(())) {
  case (f: Failure[_], _) => f
  case (s@Success(_), Success(t)) => bldr += t; s
  case (_, Failure(err)) => Failure(err)
}.map { _ => bldr.result() }
toTwitterFuture(tryRes)
```

in this way, we only go through one time.

BenFradet: Interesting, thanks a lot for your input. Will make the changes.

BenFradet:
Still investigating why the tests are failing on `multiGet`/`multiPut`.

johnynek:
@BenFradet do they fail locally as well?

BenFradet:
yup, apparently it doesn't come from the injection since I haven't seen any failure from `toTwitterFuture`.

BenFradet:
@johnynek should be fixed, I'll comment on the problems inline :+1:

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -0,0 +1,272 @@
...
+        val newKeys = result.filter(_._2.isEmpty).keySet
+
+        // handle inserts for new keys
+        val insertF =
+          if (newKeys.isEmpty) {
+            Future.Unit
+          } else {
+            // do not include None values in insert query
+            val insertKvs = newKeys.iterator
+              .collect { case k if kvs.contains(k) && kvs(k).isDefined => k -> kvs(k) }
```
BenFradet, line 196: the `&& kvs(k).isDefined` was missing compared to the original version

storehaus-mysql/src/main/scala/com/twitter/storehaus/mysql/MySqlStore.scala
```diff
@@ -0,0 +1,272 @@
...
+        val existingKvs = existingKeys.map(k => k -> kvs.getOrElse(k, None))
+
+        // do not include None values in update query
+        val updateKvs = existingKvs.iterator.collect { case (k, Some(v)) => (k, v) }.toMap
+        lazy val updateF =
+          if (updateKvs.isEmpty) Future.Unit
+          else executeMultiUpdate(updateKvs).unit
+
+        // deletes
+        val deleteKeys = existingKvs.iterator.collect { case (k, None) => k }.toSeq
```
BenFradet, line 212: the `.toSeq` was missing as a result the iterator was consumed by `deleteKeys.size` when building the sql request and `deleteParams` was empty.

johnynek (approved):

Issue #310, opened 2016-07-03
Expose Buf instead of ChannelBuffer in storehaus-redis


BenFradet:
Since Finagle is moving away from exposing netty's `ChannelBuffer` in their API in favor of twitter util's `Buf`, I think we should too.

From the recent commits, I think the replacement should be complete by Finagle's next release.

johnynek:
Yeah, I think this is a good call. (PS: lesson to be learned here, it is a good idea to hide all your dependencies on your public API in general. We didn't do that in a few places in scalding, I really regret it).

PR #328, opened 2016-09-15
Cleanup storehaus-redis

BenFradet:


BenFradet:
This will be made better once #310 is done.

storehaus-redis/src/main/scala/com/twitter/storehaus/redis/RedisHashStore.scala
```diff
@@ -45,7 +44,7 @@ class RedisHashStore(val client: Client, ttl: Option[Duration])
   extends Store[ChannelBuffer, Map[ChannelBuffer, ChannelBuffer]] {
 
   override def get(k: ChannelBuffer): Future[Option[Map[ChannelBuffer, ChannelBuffer]]] =
-    client.hGetAll(k).map({ case e if (e.isEmpty) => None case xs => Some(Map(xs:_*)) })
+    client.hGetAll(k).map { case e if e.isEmpty => None case xs => Some(Map(xs: _*)) }
```
johnynek: can we make two lines:

``` scala
.map {
  case e if e.isEmpty => None
  case xs => Some(Map(xs: _*))
}
```

johnynek (changes requested):

storehaus-redis/src/main/scala/com/twitter/storehaus/redis/RedisHashStore.scala
BenFradet: done :+1:

storehaus-redis/src/main/scala/com/twitter/storehaus/redis/RedisSortedSetStore.scala
```diff
@@ -34,7 +33,8 @@ object RedisSortedSetStore {
  */
 class RedisSortedSetStore(client: Client)
   extends MergeableStore[ChannelBuffer, Seq[(ChannelBuffer, Double)]] {
-  def semigroup = implicitly[Semigroup[Seq[(ChannelBuffer, Double)]]]
+  def semigroup: Semigroup[Seq[(ChannelBuffer, Double)]] =
+    implicitly[Semigroup[Seq[(ChannelBuffer, Double)]]]
```
johnynek, line 37: Is this really the semigroup that redis is doing? I tend to doubt it. Seems like the value is more like `Map[ChannelBuffer, Double]`. Can we add a test that the semigroup we declare here is consistent with how redis works?

We could do that by comparing `.multiMerge` here on a big set of items to what we would get from `MergeableStore.fromStore` and using the semigroup here.

BenFradet:
ping @johnynek :smile:

johnynek (commented):
sorry, I had a comment, but somehow it didn't get submitted. Maybe we don't have to fix it now, but we should file an issue, or maybe you can correct my understanding.

PR #331, opened 2016-10-25
Implementing storehaus with Postgres backend

ponkin:
Hi everybody. Here is storehouse implementation for postgres.
Need your review and comments.


storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/Read.scala
```diff
@@ -0,0 +1,15 @@
+package com.twitter.storehaus.postgres
+
+
+/**
+  * Type class for deserializing PostgreSQL native data to scala types
+  *
+  * Created by ponkin on 10/20/16.
+  */
+trait Read[A] {
```
johnynek: can we instead follow the style of the rest of the project and use `Injection[A, PostgresValue]` here instead?

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+    case x :: Nil => SELECT_SQL_PREFIX + s" = ${Show[K].show(x)}"
+    case res => SELECT_SQL_PREFIX + " IN " + res.map(Show[K].show).mkString("(", ", ", ")")
+  }
+
+  override def put(kv: (K, Option[V])): Future[Unit] = {
+    val sql = kv match {
+      case (key, Some(value)) => upsertSQL(List((key, value)))
+      case (key, None) => deleteSQL(List(key))
+    }
+    println(sql)
```
johnynek: I don't think we should have the `println` around.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+    if (ks.isEmpty) {
+      Map.empty
+    } else {
+      val result = run(selectSQL(ks.toList))
+      val reqMap = result.map {
+        _.map(row => Read[K].read(row.get(Symbol(kCol))) -> Read[V].read(row.get(Symbol(vCol)))).toMap
+      }
+      ks.foldLeft(Map.empty[K1, Future[Option[V]]]) { (acc, key) =>
+        acc + (key -> reqMap.map(_.get(key)))
+      }
```
johnynek:
``` scala
ks.iterator.map { key =>
  (key, reqMap.map(_.get(key)))
}.toMap
```

would be more direct than building the map yourself (and maybe faster since it can use a MapBuilder).

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+  private val EMPTY_STRING = ""
+
+  protected val SELECT_SQL_PREFIX =
+    s"SELECT $kCol, $vCol FROM $table WHERE $kCol"
+
+  protected val DELETE_SQL_PREFIX =
+    s"DELETE FROM $table WHERE $kCol"
+
+  override def get(key: K): Future[Option[V]] = {
+    val result = run(selectSQL(List(key)))
```
johnynek: can we pass a limit here? I'm also worried that we may have more than one item. I would rather limit to 2, then do:

``` scala
result.flatMap {
  case Nil => Future.None
  case h :: Nil => Future.value(Read[V].read(h.get(Symbol(vCol))))
  case other => Future.exception(new Exception(s"for key: $key expected 0 or 1 items, found: $other")
```

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresValue.scala
```diff
@@ -0,0 +1,14 @@
...
+import roc.postgresql.Element
+
+/**
+  * General trait to represent
+  * Postgres column value.
+  * NOTE: Actual implementation depends on driver for PostgreSQL
+  *
+  * Created by ponkin on 10/18/16.
+  */
+sealed trait PostgresValue
```
johnynek, line 13: why the extra wrapper on `Element`?

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  private def run(sql: String): Future[Result] =
+    client.query(Request(sql))
+
+  override def multiPut[K1 <: K](kvs: Map[K1, Option[V]]): Map[K1, Future[Unit]] = {
+    if (kvs.isEmpty) {
+      Map.empty
+    } else {
+      val sql = doMultiPut(kvs)
+      println(sql)
```
johnynek: let's remove this.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  private def doMultiPut[K1 <: K](kvs: Map[K1, Option[V]]): String = {
+    val (keysToUpsert, toDelete) = kvs.keySet.partition(k => kvs.get(k).exists(_.isDefined))
+    val toUpsert = kvs.filterKeys(keysToUpsert.contains).mapValues(_.get).toList
+    multiPutSQL(toUpsert, toDelete.toList)
+  }
+
+  override def close(t: Time): Future[Unit] = {
```
johnynek: don't need `{ }` for single expression methods.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,123 @@
...
+  *
+  * Created by ponkin on 9/24/16.
+  */
+object PostgresStore {
+
+  def apply[K: Show : Read, V: Show : Read](client: Client, table: String, kCol: String, vCol: String) =
+    new PostgresStore[K, V](client, table, kCol, vCol)
+}
+
+class PostgresStore[K: Show : Read, V: Show : Read]
```
johnynek: would it be too much work to add a `PostgresMergeableStore` that does `merge` in a transaction? That would be nice to have for several use cases.

johnynek (changes requested):
Thanks for the contribution. I made some suggestions. Let me know if you have any questions.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/Read.scala
ponkin: I was using Injections at first but it went me to many ugly implicit arguments like 
PostgresStore[K, V](...)(implicit inj1: Injection[K, PostgresValue], inj2: Injection[V, PostgresValue], inj3: Injection[K, Element], inj1: Injection[V, Element]) - and also expose details of implementation( Element is object from finagle-roc postgres driver). May be you can point me some example, how to do it right?

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
ponkin: Sure, I will do it!

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresValue.scala
ponkin: to change Postgres driver(finagle-roc) in future(if we we want to).

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/Read.scala
johnynek: you could do in your package:

``` scala
type PostgresValueConverter[A] = Injection[A, PostgresValue]
```

then in the `object PostgresValue { ...` define the implicits and it should be pretty clean and pick them up automatically.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresValue.scala
johnynek: I see. That's fine if we are going to automatically handle the conversions I think.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
ponkin: Well, that part turns out to be kind of tricky). How can we implement merge since we use async driver( separate calls client.query("BEGIN") and client.query("COMMIT;") can lead to unknown result, since some requests can be made simultaneously  and can be added in transaction). I do not know ho to tie Futures to particular connection. Any ideas? We can use optimistic locking but we need additional column to keep lock, like timestamp or version.

build.sbt
```diff
@@ -238,7 +238,8 @@ lazy val storehausMySQL = module("mysql").settings(
 lazy val storehausPostgres = module("postgresql").settings(
   libraryDependencies ++= Seq(
     "com.github.finagle" %% "roc-core"  % "0.0.5",
-    "com.github.finagle" %% "roc-types" % "0.0.5"
+    "com.github.finagle" %% "roc-types" % "0.0.5",
+    "com.chuusai"        %% "shapeless" % "2.3.2"
```
johnynek: do we need this? I don't see where.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/MergeablePostgresStore.scala
```diff
@@ -0,0 +1,8 @@
+package com.twitter.storehaus.postgres
+
+/**
+  * Created by ponkin on 10/26/16.
+  */
+class MergeablePostgresStore[V] {
```
johnynek: why do we need this?

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/package.scala
```diff
@@ -1,60 +1,69 @@
...
-  // Read instances
+  trait OneWayInjection[T] extends Injection[T, PostgresValue] {
+    override def apply(t: T): PostgresValue =
+      throw IllegalConversionException(s"Can`t convert ${t.getClass.getCanonicalName} to PostgresValue")
+  }
 
-  implicit val boolRead = new Read[Boolean] {
-    override def read(v: PostgresValue): Boolean = v match {
-      case RocPostgresValue(e) => e.as[Boolean]
+  implicit val boolInjection= new OneWayInjection[Boolean] {
```
johnynek, line 24: okay, I'm really sorry. I overlooked that your `Read` and `Show` were not inverses of each other. One is going to `String` the other is coming from a PostgresValue. I wish I had noticed that.

I'd actually prefer the way you had it to throwing exceptions. I mistakenly thought we could combine `Read` and `Show` into `Injection` but not the way you have the code.

What about a direct trait like:

``` scala
trait PostgresCodec[A] {
  def encode(a: A): String
  def decode(p: PostgresValue): Try[A]
}
```

I don't see a need to decouple these since they are so special purpose. I thought Injection would do, but the types are not the same on encode and decode.

johnynek (changes requested):

ponkin:
@johnynek Sorry for long delay. I added MergeablePostgresStore and tests for it. Also I rewrote all with Injections. I was trying to keep overall style. please review)

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/MergeablePostgresStore.scala
```diff
@@ -0,0 +1,69 @@
...
+    )
+  }
+
+  /**
+   * Merge two maps(merging by the key).
+   * Method will merge all values from m1 into m2.
+   * All values with the same key will be merged with Semigroup[V2].plus method
+   */
+  private def mergeWithSemigroup[K1 <: K](m1: Map[K, V], m2:  Map[K1, V]): Map[K1, V] = {
+    m2.keySet.iterator.map{ k2 =>
```
johnynek, line 62: foldLeft is generally faster because you may get some sharing:
```scala
val smaller = if (m1.size < m2.size) m1 else m2
val bigger = if (m1.size < m2.size) m2 else m1
smaller.foldLeft(bigger) { case (m, (k, v)) =>
  val newV = Semigroup.maybePlus(v, m.get(k))
  m + ((k, newV))
}
```
may need to bump the algebird dependency to get `maybePlus`, but that is fine.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  protected[postgres] def doGet[K1 <: K](k: K1)
+                               (implicit cli: Client): Future[Option[V]] =
+    doGet(Set(k)).map(_.values.headOption)
+
+  protected[postgres] def doGet[K1 <: K](ks: Set[K1],
+                                         forUpdate: Boolean = false)
+                                        (implicit cli: Client): Future[Map[K, V]] =
+    if (ks.isEmpty) {
+      Future(Map.empty)
```
johnynek, line 144: `Future.value(Map.empty)` is better since it avoids the call-by-name parameter, and since that is a constant, you might save it as a `private[this] val` and avoid the reallocation.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/package.scala
```diff
@@ -0,0 +1,79 @@
...
+  * @author Alexey Ponkin
+  */
+package object postgres {
+
+  type PostgresValueConverter[A] = Injection[A, PostgresValue]
+  type DBTxRequest[A] = Client => Future[A] // request in transaction
+
+  trait OneWayInjection[T] extends Injection[T, PostgresValue] {
+    override def apply(t: T): PostgresValue =
+      throw IllegalConversionException(
```
johnynek, line 20: can we avoid the throwing? I know we don't use it, but I really don't want to add code that throws unless we have no other way.

Here I can see two ways:

1. just add the code to wrap the primitives below in `PostgresValue`, which does not seem hard.
2. make a type-class that only goes in one direction.

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/PostgresStore.scala
```diff
@@ -0,0 +1,178 @@
...
+}
+
+class PostgresStore[K, V]
+(protected[postgres] val client: Client, table: String, kCol: String, vCol: String)
+(implicit kInj: PostgresValueConverter[K], vInj: PostgresValueConverter[V])
+  extends Store[K, V] {
+
+  private val toV: PostgresValue => V = vInj.invert(_) match {
+    case Success(v) => v
+    case Failure(e) => throw IllegalConversionException(e.getMessage)
```
johnynek, line 30: why do we need to throw here? Why not just put this in the future?
```scala
private def toV(p: PostgresValue): Future[V] = vInj.invert(p) match {
  case Success(v) => Future.value(v)
  case Failure(e) => Future.exception(e)
}
```
then use `toV(v).flatMap { ...` rather than throwing?

storehaus-postgresql/src/main/scala/com/twitter/storehaus/postgres/package.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  implicit val byteaInjection = new OneWayInjection[Array[Byte]] {
+
+    override def invert(b: PostgresValue): Try[Array[Byte]] = b match {
+      case Column(element) => Try(element.value.asInstanceOf[Array[Byte]])
+    }
+  }
+
+  def toEscString(a: Any): String = a match {
```
johnynek, line 73: I don't like this approach since we have no guarantee that `toString` is going to work. Why can't use a full injection to `PostgresValue` then use some safe mechanism to go from `Value[Any]` to `String` that can be interpolated in a SQL statement. I assume the `Value` wrapper has some support for that, no?

PR #332, opened 2016-11-03
Remove mutex from ReadThroughStore and WriteThroughStore.

ielashi:
Reasons for removing the mutex:

    1) It severely limits throughput.

    2) There is nothing about Read/WriteThroughStore that is not
       thread-safe. If it's to keep the data stores thread-safe, then it should
       be the data store's responsibility to do so.

    3) If adding the mutex is to avoid race conditions, a mutex also doesn't
       protect against that if you have multiple clients.


johnynek:
Seems reasonable, but this does not merge cleanly. Can you rebase against develop?

ielashi:
Updated to resolved conflict.

sritchie:
Works for me. Thanks @ielashi !

ielashi:
@sritchie Thanks for merging! May I request we do a release with this change?

PR #334, opened 2016-12-07
Fix for incorrect error message in MergeableStoreViaGetPut

pankajroark:
MergeableStoreViaGetPut has a bug when using FutureCollector.bestEffort. When any of the multiget requests fail they are ignored (which is correct) and afterwards all the dropped keys are mapped using missingFn which typically results in mapping to MissingValueException. The end result is that the root cause of failures of get requests is hidden and replaced with these MissingValueExceptions. This confuses users, making them think that the underlying store (such as Memcache) has lost data, whereas many times it's just timeouts. We should preserve the original exception for the get request failure, and this is what this PR is about.

But to achieve this behavior FutureCollector interface isn't sufficient, we need to get the successful results as well as the unsuccessful ones. Plus generality of using FutureCollector doesn't seem useful here. There are only two FutureCollectors currently, bestEffort and default. The default FutureCollector fails the entire multiget request if even one of the requests fails. I can't imagine any reasonable case where we'd want that behavior. So it seems best to me to remove the option of FutureCollector completely as I have proposed here. This would ultimately chain up to removing FutureCollector from MergeableStore.fromStore.

Before undertaking that api breaking change I wanted to get some feedback. Is there a valid use case for FutureCollector.default? If so any recommendations on how to organize the change. I have some ideas on how to achieve this if needed but none clean. e.g. we could fork requirements into separate functions at some level.

If the current PR looks to be on the right track I plan to remove FutureCollector for this code path and add tests for collectWithFailure.

### DIFF
--- a/storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
+++ b/storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
@@ -18,11 +18,10 @@ package com.twitter.storehaus.algebra
 
 import com.twitter.algebird.{Monoid, Semigroup}
 import com.twitter.bijection.ImplicitBijection
-import com.twitter.storehaus.{FutureCollector, FutureOps, Store}
-import com.twitter.util.{Future, Promise, Return, Throw, Try}
+import com.twitter.storehaus.{CollectionOps, FutureCollector, FutureOps, MissingValueException, Store}
+import com.twitter.util.{Future, Promise, Return, Throw}
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReferenceArray}
 import scala.language.implicitConversions
-import scala.reflect.ClassTag
 
 /** Main trait to represent stores that are used for aggregation */
 trait MergeableStore[-K, V] extends Store[K, V] with Mergeable[K, V]
@@ -57,7 +56,7 @@ object MergeableStore {
     val (collectedMGetResult, getFailures) = collectWithFailures(mGetResult)
 
     val mPutResultFut: Future[Map[K, Future[Option[V]]]] =
-      collectedMGetResult.map { pairs: Seq[(K, (Option[V], Option[V]))] =>
+      collectedMGetResult.map { pairs: Array[(K, (Option[V], Option[V]))] =>
         val pairMap = pairs.toMap
         val mPutResult: Map[K, Future[Unit]] = store.multiPut(pairMap.mapValues(_._2))
         mPutResult.map { case (k, funit) =>
@@ -65,31 +64,38 @@ object MergeableStore {
         }
       }
 
+    val missingFn: K => Future[Option[V]] = { k =>
+      getFailures.flatMap { failures =>
+        Future.exception(failures.getOrElse(k, new MissingValueException[K](k)))
+      }
+    }
+
     /**
-     * Combine original keys with result after put.
+     * Combine original keys with result after put and errors with get.
      */
-    FutureOps.liftFutureValues(keySet, mPutResultFut, getFailures)
+    FutureOps.liftFutureValues(keySet, mPutResultFut, missingFn)
   }
 
   /**
    * Collects keyed futures, partitioning out the failures.
    */
-  private[algebra] def collectWithFailures[K, V](fs:Map[K, Future[V]]): (Future[Seq[(K, V)]], Map[K, Future[Nothing]]) = {
+  private[algebra] def collectWithFailures[K, V](fs: Map[K, Future[V]]): (Future[Array[(K, V)]], Future[Map[K, Throwable]]) = {
     if (fs.isEmpty) {
-      (Future.value(Seq.empty[(K, V)]), Map.empty[K, Future[Nothing]])
+      (Future.value(Array.empty[(K, V)]), Future.value(Map.empty[K, Throwable]))
     } else {
       val fsSize = fs.size
-      val results = new AtomicReferenceArray[Either[(K,Future[Nothing]), (K, V)]](fsSize)
+      val results = new AtomicReferenceArray[Either[(K, Throwable), (K, V)]](fsSize)
       val countdown = new AtomicInteger(fsSize)
       val successCount = new AtomicInteger(0)
-      val pSuccess = new Promise[Seq[(K, V)]]
-      var failures = Map.empty[K, Future[Nothing]]
+      val pSuccess = new Promise[Array[(K, V)]]
+      val pFailures = new Promise[Map[K, Throwable]]
 
       def collectResults() = {
         if (countdown.decrementAndGet() == 0) {
           val successArray = new Array[(K, V)](successCount.get())
           var si = 0
           var ri = 0
+          var failures = Map.empty[K, Throwable]
           while (ri < fsSize) {
             results.get(ri) match {
               case Right(kv) =>
@@ -101,6 +107,7 @@ object MergeableStore {
             ri += 1
           }
           pSuccess.setValue(successArray)
+          pFailures.setValue(failures)
         }
       }
 
@@ -111,12 +118,12 @@ object MergeableStore {
             successCount.incrementAndGet()
             collectResults()
           case t@Throw(cause) =>
-            val failure = k -> Future.const(Throw(cause))
+            val failure = k -> cause
             results.set(i, Left(failure))
             collectResults()
         }
       }
-      (pSuccess, failures)
+      (pSuccess, pFailures)
     }
   }
 
--- a/storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
+++ b/storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
@@ -18,7 +18,7 @@ package com.twitter.storehaus.algebra
 
 import com.twitter.algebird.{Monoid, Semigroup}
 import com.twitter.bijection.ImplicitBijection
-import com.twitter.storehaus.{FutureCollector, FutureOps, MissingValueException, Store}
+import com.twitter.storehaus.{CollectionOps, FutureCollector, FutureOps, MissingValueException, Store}
 import com.twitter.util.{Future, Promise, Return, Throw, Try}
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReferenceArray}
 import scala.language.implicitConversions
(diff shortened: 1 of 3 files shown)


storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -31,25 +32,29 @@ object MergeableStore {
   implicit def enrich[K, V](store: MergeableStore[K, V]): EnrichedMergeableStore[K, V] =
     new EnrichedMergeableStore(store)
 
+  private[this] def addOpt[V](init: Option[V], inc: V)(implicit sg: Semigroup[V]): Option[V] = init match {
```
johnynek, line 34: 1. the latest algebird has `Semigroup.maybePlus` this does this, but it always returns `V` not `Option[V]` which I think is correct (you can just box in `Some` as needed).

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -61,10 +66,58 @@ object MergeableStore {
...
+     * Combine original keys with result after put.
      */
-    FutureOps.liftFutureValues(keySet, mPutResultFut, missingfn)
+    FutureOps.liftFutureValues(keySet, mPutResultFut, getFailures)
+  }
+
+  /**
+   * Collects keyed futures, partitioning out the failures.
+   */
+  private[algebra] def collectWithFailures[K, V](fs:Map[K, Future[V]]): (Future[Seq[(K, V)]], Map[K, Future[Nothing]]) = {
```
johnynek: can we put spaces in types: `fs: Map...`

johnynek: I'm not crazy about using `Seq`. I'd rather use `List` or `Vector` so the contracts are clearer.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -61,10 +66,58 @@ object MergeableStore {
...
+    if (fs.isEmpty) {
+      (Future.value(Seq.empty[(K, V)]), Map.empty[K, Future[Nothing]])
+    } else {
+      val fsSize = fs.size
+      val results = new AtomicReferenceArray[Either[(K,Future[Nothing]), (K, V)]](fsSize)
```
johnynek: what about `AtomicReferenceArray[(K, Either[Throwable, V])]` here instead. Seems easier to follow

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
johnynek: If this is to be private, it can even be `Array`.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -61,10 +66,58 @@ object MergeableStore {
...
+      val countdown = new AtomicInteger(fsSize)
+      val successCount = new AtomicInteger(0)
+      val pSuccess = new Promise[Seq[(K, V)]]
+      var failures = Map.empty[K, Future[Nothing]]
```
johnynek: doesn't this need to be an `AtomicReference`? How is it safe to modify this var from many threads?

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -61,10 +66,58 @@ object MergeableStore {
...
+            results.set(i, Right(k -> v))
+            successCount.incrementAndGet()
+            collectResults()
+          case t@Throw(cause) =>
+            val failure = k -> Future.const(Throw(cause))
+            results.set(i, Left(failure))
+            collectResults()
+        }
+      }
+      (pSuccess, failures)
```
johnynek: failures has to be a future right? You are returning a copy here, but the futures are not done yet. If you could have a `Map` you could see what keys have failed, and you can't can you? I think this has to be `Future[Map[K, ..]]`. I am not sure `Future[Nothing]` is a better type that `Throwable` though.

johnynek (changes requested):

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: Makes sense. Don't want to merge latest algebird in this change. Is it ok to simplify this separate review where I can merge algebird. I've left a todo for now.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: fixed

pankajroark: Replaced with Array.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: Yes, that was totally wrong. I've fixed it now and added tests as well.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: Done

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -32,6 +31,7 @@ object MergeableStore {
   implicit def enrich[K, V](store: MergeableStore[K, V]): EnrichedMergeableStore[K, V] =
     new EnrichedMergeableStore(store)
 
+  // todo(pankajg) After merging latest algebird replace this with Semigroup.maybePlus
```
johnynek: can we not use `todo` and instead file issues. We don't always close issues, but we approximately never address todos.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -57,66 +57,73 @@ object MergeableStore {
...
      */
-    FutureOps.liftFutureValues(keySet, mPutResultFut, getFailures)
+    FutureOps.liftFutureValues(keySet, mPutResultFut, missingFn)
   }
 
   /**
    * Collects keyed futures, partitioning out the failures.
    */
-  private[algebra] def collectWithFailures[K, V](fs:Map[K, Future[V]]): (Future[Seq[(K, V)]], Map[K, Future[Nothing]]) = {
+  private[algebra] def collectWithFailures[K, V](fs: Map[K, Future[V]]): (Future[Array[(K, V)]], Future[Map[K, Throwable]]) = {
```
johnynek: looking at this again it seems like rather than `Either[Throwable, V]` we could just use `Try[V]` which is basically the same.

Then this method signature could be:
`def collectWithFailures[K, V](fs: Map[K, Future[V]]): Future[Map[K, Try[V]]]` which actually seems useful enough to have even in twitter util.

In any case, the first thing you do when you get the result is call `.toMap` on the array, so returning the `Map` seems to make sense.

Also, both of these futures are satisfied at the same time, right? So, even if you don't merge the data structures, you could do: `Future[(Map[K, V], Map[K, Throwable])]` so it is clear that all the data comes in at one time.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
```diff
@@ -57,66 +57,73 @@ object MergeableStore {
...
+                failures = failures + (k -> t)
             }
             ri += 1
           }
           pSuccess.setValue(successArray)
+          pFailures.setValue(failures)
         }
       }
 
       for (((k, f), i) <- fs.iterator.zipWithIndex) {
```
johnynek: if you are going to geek out on while loops above, doing the same here is probably worth it:
```scala
val mapIt = fs.iterator
var i = 0
while(mapIt.hasNext) {
  val kv = mapIt.next
  fv._2.respond {
    case Return(v) =>...
  }
  i += 0
}
```

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: The idea of Either[Throwable, V] to Try[V] is great, indeed Try is basically the same in this context and simpler.

Combining successful and failed results will be inefficient since we need to immediately do multiput for the successful ones and so need them separately. By passing them both back in a single future we have to split them up again immediately after. I do agree that returning a single future conveys the behavior more correctly. It feels like a question of performance and code simplicity vs indication of correct intent. I've implemented the return with single future PTAL.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: Implemented this, it does add a bit of complexity. Also does avoid one wrapping/unwraping for each item in the map.

storehaus-algebra/src/main/scala/com/twitter/storehaus/algebra/MergeableStore.scala
pankajroark: Sounds good. Not sure if filing a jira before the change is merged would be a good idea. Will file the jira after this change merges. This is a bit of problem with filing jiras in such cases, it is something that would need to be kept track separately.

pankajroark:
I've removed couple of intermediate map creations. In this case I think optimizing is worthwhile because I've seen this function show up in YourKit profiles as taking noticeable cpu. Plus the extra maps will create garbage.

johnynek:
:+1: when tests are green. Nice result. Good work removing more intermediate Map creations.

pankajroark:
So, the binary compatibility tests are failing because because the signature of the function has changed. This is another thing I wanted feedback on, that it is ok to make this binary incompatible change. Also, how do I make the tests pass in case of binary incompatible changes, by incrementing version?

sritchie-stripe:
Hey @pankajroark,

I'd say it's fine for now since we have so damned many binary incompatible changes stacked up. To make the tests pass, copy the suggested line from the mima check in the build into here:

https://github.com/twitter/storehaus/blob/develop/build.sbt#L31

Once that's done we'll merge!

johnynek:
can't you implement the old function still? Do we need to add the binary incompat?

To me, each binary (and source) incompatible change puts a cost out on users, and slows adoption (people get stuck on old versions rather than pay the tax to move up).

If it it easy (or only slightly a pain), we should keep source and binary compatibility.

In this case, it is easy to add a second method where the user passes a function in, right? you are basically inlining the old default on line 75, right?

pankajroark:
MissingValueException there will actually never hit, I kept it there just
for safety. Letting users supply that would be misleading. Also we don't
use the future collector. FutureCollector doesn't allow capturing
individual failures.

On Thu, Dec 15, 2016 at 1:13 PM P. Oscar Boykin <notifications@github.com>
wrote:

can't you implement the old function still? Do we need to add the binary
incompat?

To me, each binary (and source) incompatible change puts a cost out on
users, and slows adoption (people get stuck on old versions rather than pay
the tax to move up).

If it it easy (or only slightly a pain), we should keep source and binary
compatibility.

In this case, it is easy to add a second method where the user passes a
function in, right? you are basically inlining the old default on line 75,
right?

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/twitter/storehaus/pull/334#issuecomment-267445579>, or mute
the thread

johnynek:
okay, if you are fairly sure that the old API does not make sense anymore, I guess. But adding a @deprecated and calling the new API can help users other than yourself.

For instance, we get stuck on all kinds of old versions at Stripe over little issues like this. Each one may not be a big deal, but when you try to bump summingbird and you have to bump util, finagle, bijection, algebird, scalding and storehaus, and each has a few incompatibilities, it turns into a really big job.

We can add this method to this list:

to get the tests green (keeping in mind we should do a version bump next and remove all that list).

pankajroark:
Thanks for the guidance, will do that.

johnynek (approved):

johnynek:
👍

PR #336, opened 2016-12-11
Convert storehaus-caliper -> storehaus-benchmark w/ JMH

sritchie:
This PR removes the caliper dependency and subs in JMH, to make this project more consistent with our other projects.


johnynek:
👍

PR #342, opened 2017-02-21
Preview of upgrade to Scala 2.12.

ghost:
1. Incremented all required versions to value that included scala 2.12 support. See diff for details.
2. Modified RedisStore to use Buf instead of ChannelBuffer due to changes in underlying client. To ease this transition a method to convert Store[Buf, Buf] to Store[ChannelBuffer, ChannelBuffer] is provided on RedisStore.
3. Removed support for Scala 2.10. Two libraries lacked support for scala 2.10 and 2.12 on the same version: twiiter-util and finagle. Supporting 2.10 finagle and 2.12 finagle was not possible due to interface changes.
4. Various packages changed and were fixed.
5. Added target to travis.yml and build.sbt to include scala 2.12.
6. Set source and build target to java 8 to support scala 2.12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,24 @@ language: scala
 sudo: false
 matrix:
   include:
-    - scala: 2.10.6
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues
-
     - scala: 2.11.8
+      jdk: oraclejdk8
       script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
       after_success:
         - bash <(curl -s https://codecov.io/bash)
+#    - scala: 2.12.1
+#      jdk: oraclejdk8
+#      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+#      after_success:
+#        - bash <(curl -s https://codecov.io/bash)
 before_script:
+  - sleep 15 # https://docs.travis-ci.com/user/database-setup/#MongoDB
+  - mongo mydb_test --eval 'db.addUser("travis", "test");'
   - mysql -u root -e "create database storehaus_test;"
   - mysql -u root -e "create user 'storehaususer'@'localhost' identified by 'test1234';"
   - mysql -u root -e "grant all on storehaus_test.* to 'storehaususer'@'localhost';"
 services:
+  - mysql
   - redis-server
   - memcache
   - mongodb
--- a/build.sbt
+++ b/build.sbt
@@ -65,9 +65,9 @@ val ignoredABIProblems = {
 val sharedSettings = extraSettings ++ ciSettings ++ Seq(
   organization := "com.twitter",
   scalaVersion := "2.11.7",
-  crossScalaVersions := Seq("2.10.6", "2.11.7"),
-  javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
-  javacOptions in doc := Seq("-source", "1.6"),
+  crossScalaVersions := Seq("2.11.7", "2.12.1"),
+  javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
+  javacOptions in doc := Seq("-source", "1.8"),
   libraryDependencies += "org.scalatest" %% "scalatest" % scalatestVersion % "test",
   resolvers ++= Seq(
     Opts.resolver.sonatypeSnapshots,
@@ -86,7 +86,7 @@ val sharedSettings = extraSettings ++ ciSettings ++ Seq(
   ),
 
   // add linter for common scala issues: https://github.com/HairyFotr/linter
-  addCompilerPlugin("org.psywerx.hairyfotr" %% "linter" % "0.1.14"),
+  addCompilerPlugin("org.psywerx.hairyfotr" %% "linter" % "0.1.17"),
 
   // Publishing options:
 
@@ -162,12 +162,14 @@ lazy val noPublishSettings = Seq(
     publishArtifact := false
   )
 
-val algebirdVersion = "0.12.0"
-val bijectionVersion = "0.9.1"
-val utilVersion = "6.34.0"
-val scaldingVersion = "0.16.0-RC1"
-val finagleVersion = "6.35.0"
-val scalatestVersion = "2.2.4"
+val algebirdVersion = "0.13.0"
+val bijectionVersion = "0.9.5"
+val utilVersion = "6.39.0"
+
+val scaldingVersion = "0.17.0"
+val finagleVersion = "6.41.0"
+val scalatestVersion = "3.0.1"
+val scalaCheckVersion = "1.13.4"
 
 lazy val storehaus = Project(
   id = "storehaus",
@@ -228,9 +230,8 @@ lazy val storehausMemcache = module("memcache").settings(
     "com.twitter" %% "bijection-core" % bijectionVersion,
     "com.twitter" %% "bijection-netty" % bijectionVersion,
     "com.twitter" %% "finagle-memcached" % finagleVersion excludeAll(
-      // we don't use this and its not on maven central.
-      ExclusionRule("com.twitter.common.zookeeper"),
-      ExclusionRule("com.twitter.common")
+      // we don't use this
+      ExclusionRule("com.twitter.common.zookeeper")
       )
   )
 ).dependsOn(storehausAlgebra % "test->test;compile->compile")
@@ -290,7 +291,7 @@ lazy val storehausKafka = module("kafka").settings(
     "com.twitter" %% "bijection-avro" % bijectionVersion,
     "org.apache.kafka" % "kafka-clients" % "0.9.0.1",
     "org.apache.zookeeper" % "zookeeper" % "3.4.8" % "test",
-    "org.apache.kafka" %% "kafka" % "0.9.0.1" % "test"
+    "org.apache.kafka" %% "kafka" % "0.10.1.1" % "test"
   ),
   // we don't want various tests clobbering each others keys
   parallelExecution in Test := false
@@ -299,7 +300,7 @@ lazy val storehausKafka = module("kafka").settings(
 lazy val storehausMongoDB = module("mongodb").settings(
   libraryDependencies ++= Seq(
     "com.twitter" %% "bijection-core" % bijectionVersion,
-    "org.mongodb" %% "casbah" % "2.8.2"
+    "org.mongodb" %% "casbah" % "3.1.1"
   ),
   parallelExecution in Test := false
 ).dependsOn(storehausAlgebra % "test->test;compile->compile")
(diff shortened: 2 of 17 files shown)


ghost:
I had some trouble getting the tests to run consistently on travis.ci. When I ran the tests locally against each data-store running in docker everything worked. 

Noticed the mongo-db tests create ~10k connections during the test run this seemed to be part of why travis has trouble with the build.

BenFradet:
Nice, this would solve #310 as well :+1:

build.sbt
```diff
@@ -227,9 +230,8 @@ lazy val storehausMemcache = module("memcache").settings(
     "com.twitter" %% "bijection-core" % bijectionVersion,
     "com.twitter" %% "bijection-netty" % bijectionVersion,
     "com.twitter" %% "finagle-memcached" % finagleVersion excludeAll(
-      // we don't use this and its not on maven central.
-      ExclusionRule("com.twitter.common.zookeeper"),
-      ExclusionRule("com.twitter.common")
```
ghost, line 233: Part of the upgraded finagle memcache client needed com.twitter.common.metrics.Gauge so this exclude was removed.

build.sbt
```diff
@@ -161,12 +161,15 @@ lazy val noPublishSettings = Seq(
     publishArtifact := false
   )
 
-val algebirdVersion = "0.12.0"
-val bijectionVersion = "0.9.1"
-val utilVersion = "6.34.0"
+val algebirdVersion = "0.12.4"
```
johnynek: this should be `0.13.0`

build.sbt
```diff
@@ -322,7 +324,7 @@ lazy val storehausTesting = Project(
   settings = sharedSettings ++ Seq(
     name := "storehaus-testing",
     libraryDependencies ++= Seq(
-      "org.scalacheck" %% "scalacheck" % "1.12.2" withSources(),
+      "org.scalacheck" %% "scalacheck" % "1.12.6" withSources(),
```
johnynek: can we go to `1.13` branch? Also, while we are at it, can we move this version up to the top of the file where we set the other versions?

ghost:
@johnynek is the general preference to be on the max version of each lib? I set the versions to the min  available.

johnynek:
The general preference is to be on the minimum version that gives us what we need, that is also binary compatible with the latest version. Consumers of this library can then set.

For scalacheck, 1.13 has some significantly better features (better Function generation, and I thought was needed for 2.12).

ghost:
Great. Thanks for that guidance. I will go back over the versions set here to make sure everything is aligned with that.

ghost:
Looking at the travis ci failure it might be that the level of concurrency is too high for mysql and the other stores to handle during these property tests.

piyushnarang:
@ben-adazza - Published the 2.12 friendly version of Scalding - 0.17.0. Can you try and pick that up in your PR?

build.sbt
```diff
@@ -161,12 +161,16 @@ lazy val noPublishSettings = Seq(
...
-val algebirdVersion = "0.12.0"
-val bijectionVersion = "0.9.1"
-val utilVersion = "6.34.0"
+val algebirdVersion = "0.13.0"
+val bijectionVersion = "0.9.5"
+val utilVersion = "6.39.0"
+
+//Scalding is not yet on  scala 12 so this needs to be
+// rev'ed to get the PR to compile for 12.
 val scaldingVersion = "0.16.0-RC1"
```
piyushnarang: can upgrade to 0.17.0

ghost: Excellent. I will do that right away. Thanks!

.travis.yml
```diff
@@ -2,18 +2,24 @@ language: scala
...
   include:
-    - scala: 2.10.6
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues
-
     - scala: 2.11.8
+      jdk: oraclejdk8
       script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
       after_success:
         - bash <(curl -s https://codecov.io/bash)
+#    - scala: 2.12.1
```
piyushnarang, line 10: we can uncomment this now right?

piyushnarang: I think we dont need to run coverage (as we're doing this for 2.11) and mima (as this will be the first 2.12 build).

piyushnarang:
Took a quick look and it seems like a bunch of our mysql tests are failing due to:
```
WARNING: Exception propagated to the default monitor (upstream address: n/a, downstream address: /127.0.0.1:3306, label: 127.0.0.1:3306).
com.twitter.finagle.mysql.ServerError: Too many connections
	at com.twitter.finagle.mysql.ClientDispatcher$$anonfun$com$twitter$finagle$mysql$ClientDispatcher$$decodePacket$6.apply(ClientDispatcher.scala:228)
	at com.twitter.finagle.mysql.ClientDispatcher$$anonfun$com$twitter$finagle$mysql$ClientDispatcher$$decodePacket$6.apply(ClientDispatcher.scala:226)
	at com.twitter.util.Future$$anonfun$flatMap$1.apply(Future.scala:1089)
	at com.twitter.util.Future$$anonfun$flatMap$1.apply(Future.scala:1088)
...
```

ghost:
Yeah. I think the properties tests create 10s of thousands on of connections since they throw away an entire client after each test.

piyushnarang:
@ben-adazza yeah we either need to update them to close the connection or set the mysql connection limits higher - https://dev.mysql.com/doc/refman/5.7/en/too-many-connections.html
Curious as to why this was not failing earlier

ghost:
Likely a change to the underlying implementation in finagle-mysql.

build.sbt
```diff
@@ -162,12 +162,14 @@ lazy val noPublishSettings = Seq(
...
-val utilVersion = "6.34.0"
-val scaldingVersion = "0.16.0-RC1"
-val finagleVersion = "6.35.0"
-val scalatestVersion = "2.2.4"
+val algebirdVersion = "0.13.0"
+val bijectionVersion = "0.9.5"
+val utilVersion = "6.39.0"
+
+val scaldingVersion = "0.17.0"
+val finagleVersion = "6.41.0"
```
piyushnarang, line 170: think the latest version is 6.43.0. Let's use that?

piyushnarang:
@ben-adazza so looking at the example listed in the finagle-mysql repo: https://github.com/twitter/finagle/blob/master/finagle-example/src/main/scala/com/twitter/finagle/example/mysql/Example.scala#L45
And then looking at our code, https://github.com/twitter/storehaus/blob/develop/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlStoreProperties.scala#L141 it seems as though we're setting things up to call close. I tried repro-ing this on my machine but my mysql setup is horked. Are you able to repro and see if the close is being called / not?

pankajroark:
I'm able to repro the mysql test failure on my machine. Something that worked for me is sharing a client across tests. But client is not thread safe, so I also had to put the invocations in a synchronized block. Here's the patch:
```
diff --git a/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlLongStoreProperties.scala b/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlLongStoreProperties.scala
index 378e8bb..c2ab74b 100644
--- a/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlLongStoreProperties.scala
+++ b/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlLongStoreProperties.scala
@@ -89,6 +89,12 @@ object MySqlLongStoreProperties extends Properties("MySqlLongStore")
     isMatch
   }
 
+  // these should match mysql setup used in .travis.yml
+  lazy val client = Mysql.client
+    .withCredentials("storehaususer", "test1234")
+    .withDatabase("storehaus_test")
+    .newRichClient("127.0.0.1:3306")
+
   property("MySqlLongStore put and get") =
     withStore(putAndGetStoreTest(_), "text", "bigint")
 
@@ -97,18 +103,15 @@ object MySqlLongStoreProperties extends Properties("MySqlLongStore")
 
   private def withStore[T](f: MySqlLongStore => T, kColType: String, vColType: String,
       merge: Boolean = false): T = {
-    val client = Mysql.client
-      .withCredentials("storehaususer", "test1234")
-      .withDatabase("storehaus_test")
-      .newRichClient("127.0.0.1:3306")
-    // these should match mysql setup used in .travis.yml
 
     val tableName = s"storehaus-mysql-long-$kColType-$vColType${if (merge) "-merge" else ""}"
     val schema = s"CREATE TEMPORARY TABLE IF NOT EXISTS `$tableName` (`key` $kColType DEFAULT " +
       s"NULL, `value` $vColType DEFAULT NULL) ENGINE=InnoDB DEFAULT CHARSET=utf8;"
-    Await.result(client.query(schema))
+    this.synchronized {
+      Await.result(client.query(schema))
+      f(newStore(client, tableName))
+    }
 
-    f(newStore(client, tableName))
   }
 
   def newStore(client: Client, tableName: String): MySqlLongStore =
diff --git a/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlStoreProperties.scala b/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlStoreProperties.scala
index 26633ff..c192bc8 100644
--- a/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlStoreProperties.scala
+++ b/storehaus-mysql/src/test/scala/com/twitter/storehaus/mysql/MySqlStoreProperties.scala
@@ -99,6 +99,12 @@ object MySqlStoreProperties extends Properties("MySqlStore")
       s"expected value $expected, but found $found")
   }
 
+  // these should match mysql setup used in .travis.yml
+  lazy val client = Mysql.client
+    .withCredentials("storehaususer", "test1234")
+    .withDatabase("storehaus_test")
+    .newRichClient("127.0.0.1:3306")
+
   property("MySqlStore text->text") =
     withStore(putAndGetStoreTest(_), "text", "text")
 
@@ -140,17 +146,14 @@ object MySqlStoreProperties extends Properties("MySqlStore")
 
   private def withStore[T](f: MySqlStore => T, kColType: String, vColType: String,
       multiGet: Boolean = false): T = {
-    val client = Mysql.client
-      .withCredentials("storehaususer", "test1234")
-      .withDatabase("storehaus_test")
-      .newRichClient("127.0.0.1:3306")
-    // these should match mysql setup used in .travis.yml
 
     val tableName = s"storehaus-mysql-$kColType-$vColType${if (multiGet) "-multiget" else ""}"
     val schema = s"CREATE TEMPORARY TABLE IF NOT EXISTS `$tableName` (`key` $kColType " +
       s"DEFAULT NULL, `value` $vColType DEFAULT NULL) ENGINE=InnoDB DEFAULT CHARSET=utf8;"
-    Await.result(client.query(schema))
-    f(newStore(client, tableName))
+    this.synchronized {
+      Await.result(client.query(schema))
+      f(newStore(client, tableName))
+    }
   }
 
   def newStore(client: Client, tableName: String): MySqlStore =
```

pankajroark:
@ben-adazza btw are you still interested in pursuing this PR? I need to upgrade storehaus to scala 2.12 urgently, so I'm happy to take over this work in another PR if needed. Of course really appreciate the work you've already done. Happy to help in any way to get this upgrade shipped.

ghost:
@pankajroark by all means you can take this code and get the tests running. Debugging these tests is not high on my priority list right now.

pankajroark:
Fyi: the thing that directly breaks the mysql test is scalacheck update. But the fundamental problem is that the test doesn't close the client. Client can be closed only after the property is used in the tests. I'm trying to find a hook that executes after a property is done testing, to put the client close code there.

ghost:
This work was very kindly taken over in: https://github.com/twitter/storehaus/pull/346/files

PR #343, opened 2017-03-15
Occupancy for Cache trait implementation classes

ppanero:
Added a function to be able to access the cache occupancy (amount of elements) of it


ppanero:
I noticed it fails with
```
method occupancy()Int in trait com.twitter.storehaus.cache.Cache is present only in current version
```
But im not sure how to fix it.

johnynek:
To fix the build, add the method to this list:

This looks fine to me. I think there is no way around breaking binary compatibility if we want this method, and it looks like we have a lot of pending breakages.

If you get this green, I'm happy to merge.

ppanero:
Had some issues in my branches etc. Closed this, squashed and PR again: https://github.com/twitter/storehaus/pull/345

Cheers

PR #344, opened 2017-03-16
Upgrate to Finagle 6.43.0 and preparing the path for Scala 2.12

csaltos:
The main goal of this PR is to update Finagle to the new version 6.43.0 for using new and improve Finagle features and performance and also one day migrate Storehaus to Scala 2.12.

The **Scala 2.10 cross build is still kept valid** using Finagle 6.35.0 (new Finagle versions beyond 6.35.0 are not compatible with Scala 2.10 any more which includes, of course Finagle 6.43.0)

Scala 2.12 is not added to the cross builds yet, since the **Scalding dependency is not yet available for Scala 2.12** (all the other dependencies are already Scala 2.12 compatible updating their versions)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,17 +56,39 @@ val ignoredABIProblems = {
       ".futurePool"),
     exclude[MissingMethodProblem]("com.twitter.storehaus.asynchbase.AsyncHBaseStore.futurePool"),
     exclude[MissingMethodProblem]("com.twitter.storehaus.ReadThroughStore.mutex"),
-    exclude[MissingClassProblem]("com.twitter.storehaus.kafka.JavaFutureToTwitterFutureConverter$Closed$"),
-    exclude[DirectMissingMethodProblem]("com.twitter.storehaus.kafka.KafkaStore.<init>$default$3")
+    exclude[MissingClassProblem]("com.twitter.storehaus.kafka" +
+      ".JavaFutureToTwitterFutureConverter$Closed$"),
+    exclude[DirectMissingMethodProblem]("com.twitter.storehaus.kafka.KafkaStore.<init>$default$3"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.http.HttpException.apply"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.MySqlStore.apply"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.MySqlLongStore.apply"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.ValueMapper" +
+      ".toChannelBuffer"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.ValueMapper.toLong"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.ValueMapper.toString"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.storehaus.mysql.MySqlValue.v"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.storehaus.mysql.MySqlValue.this"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.storehaus.mysql.MySqlStore.client"),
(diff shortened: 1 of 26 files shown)


csaltos:
The style of the code, name of the classes and packages or folders can be changed to keep the original convention the project uses, of course.

Let me know if a change, adjustment or fix this PR requires to be accepted (at Talenteca.com we would love to have a newer Storehaus with Finagle 6.43.0 available)

johnynek:
looks like we need to update the `.travis.yml` to add 2.12 to the tests and use jdk8 for the 2.11 branch.

Thanks for taking the time to keep the 2.10 build green. I appreciate you doing that.

johnynek:
how does this relate to #342? It seems like we might want to merge this first and leverage the work here in that one?

csaltos:
The #342 eliminates the support for Scala 2.10 while this one keeps the Scala 2.10 and also upgrades Finagle to a newer version for Scala 2.11 (and even Scala 2.12 when we're ready to do it).

The #342 is already updating Storehaus to Scala 2.12 so we can merge those parts from that PR (once the Scalding part is resolved because that's still pending since Scalding is not Scala 2.12 compatible yet)

Thank you for taking a look at Travis, if I can be of any help on that part, let me know.

ben-adazza:
The travis build is failing here:  `java.lang.UnsupportedClassVersionError: com/twitter/util/Duration$ : Unsupported major.minor version 52.0`. This is because some of jars in the new lib versions being used were compiled under java 8. I fixed this in https://github.com/twitter/storehaus/pull/342/ by setting the target source and compile to 8. 

To continue to use scala 10 you would need to modify the sbt config to use different source targets per scala version I think.

csaltos:
Great, thank you @ben-adazza. I'll try to fix this PR following your advice.

csaltos:
Ready @ben-adazza and @johnynek, this PR is now building successfully with travis CI and working for Scala 2.10 and Scala 2.11 for the new Finagle version (with this PR we get closer and closer to upgrade to Scala 2.12 soon)

pankajroark:
Curios what this PR is waiting on. @johnynek @ben-adazza is this ok to merge? Asking because there is some overlap with the other PR around scala 2.12 upgrade.

build.sbt
```diff
@@ -164,9 +185,9 @@ lazy val noPublishSettings = Seq(
 
 val algebirdVersion = "0.12.0"
 val bijectionVersion = "0.9.1"
-val utilVersion = "6.34.0"
+def utilVersion(scalaVersionValue: String) = if (scalaVersionValue.startsWith("2.10.")) "6.34.0" else "6.42.0"
```
pankajroark, line 188: Not sure how important it is to support 2.10. Having two versions of external dependencies makes maintenance harder.

storehaus-http/src/main/scala-2.10/com/twitter/storehaus/http/compat/package.scala
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twitter.storehaus.http
+
+package object compat {
```
pankajroark, line 19: Again, not sure how important it is to support 2.10. If not then looks like we can avoid this compat code.

pankajroark (commented):
Looks like a good change. But I'm concerned about the additional complexity we are adding to support scala 2.10. Do we really need to support scala 2.10, otherwise we can simplify things a lot.

storehaus-http/src/main/scala-2.10/com/twitter/storehaus/http/compat/package.scala
csaltos: Of course making the decision to not to support Scala 2.10 any more will make maintenace easier.

I see that #343 just gets approbed so this present PR (#344) does not makes any sense now.

Thank you for upgrading  storehaus to Scala 2.12 and a newer Finagle for the next storehaus version (in our case we don't use Scala 2.10 at our production real servers so it's OK, so the new storehaus is very valuable to us, as always, thanks for that).

PR #347, opened 2017-05-11
Upgrade finagle to 6.43.0

pankajroark:
Upgrades finagle and twitter util. In the process gets rid of finagle-http-compat and moves to the latest finagle http api.

### DIFF
--- a/storehaus-http/src/main/scala/com/twitter/storehaus/http/HttpStore.scala
+++ b/storehaus-http/src/main/scala/com/twitter/storehaus/http/HttpStore.scala
@@ -40,9 +40,10 @@ case class HttpException(code: Int, reasonPhrase: String, content: String)
 object HttpStore {
   def apply(dest: String): HttpStore = new HttpStore(Http.newService(dest))
 
-  def toChannelBufferStore(store: Store[Buf, Buf]) : Store[ChannelBuffer, ChannelBuffer] = {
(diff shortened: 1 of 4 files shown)


storehaus-http/src/main/scala/com/twitter/storehaus/http/HttpStore.scala
```diff
@@ -17,72 +17,89 @@
...
+  override def put(kv: (String, Option[Buf])): Future[Unit] = {
     val request = kv match {
-      case (k, Some(cb)) =>
-        val req = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.PUT, k)
-        req.setContent(cb)
-        req.headers.set(HttpHeaders.Names.CONTENT_LENGTH, cb.readableBytes.toString)
+      case (k, Some(buf)) =>
+        val req = Request(Method.Put, k)
+        req.content = buf
+        req.contentLength = buf.length
```
ttim, line 75: Is `buf.length` the same as `.readableBytes`?

ttim (approved):
Looks good!

storehaus-http/src/main/scala/com/twitter/storehaus/http/HttpStore.scala
pankajroark: yeah, seem to be afaict.

PR #346, opened 2017-05-07
Scala 2.12 upgrade preview with mysql tests fixed

pankajroark:
Builds on top of #342 fixing the mysql tests that the scalacheck upgrade breaks.

PR #348, opened 2017-05-12
Release 0.15.0 prep. Update version in preparation for rc release.

pankajroark:
Proposed release notes:
* Add support for scala 2.12 #346 
* Update finagle to 6.43.0 #347 

Note that this release drops support for anything before jdk8. Upstream libs finagle and util have already done so.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.15.0-SNAPSHOT"
+version in ThisBuild := "0.15.0-RC2"
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.15.0-RC2"
+version in ThisBuild := "0.15.0"


johnynek:
seems good to me. We can also go ahead and reset the mima to not check.

Also, the next version will be 0.15.0, not 0.16.0.

pankajroark:
Can you clarify about "We can also go ahead and reset the mima to not check." I'm not familiar with the procedure around mima.

johnynek:
this list should be emptied out:

and this should be set to `0.15.0` after we publish.

ttim:
@johnynek I thought we do next (at least I did with tormenta):
1. set version to `0.15.0`, do release, point to this commit as release commit
2. set version to `0.15.1-snapshot`, clean mima exclusions, change `youngestForwardCompatible` to point to `0.15.0` as subsequent PR after release.

But for tormenta we had actual release tags (and we set version in this tags with later merge them into develop).
What do you think about this workflow? Should it be somehow different?

johnynek:
I think your workflow is fine an less likely to cause an issue. I was assuming no one changes any code after dropping the exclusions, but indeed it is better to keep them there until we can bump the version to `0.15.1.-snapshot`.

pankajroark:
published 0.15.0-RC2, and updated the version in PR to 0.15.0. Waiting for this to be accepted to merge, then tag commit for 0.15.0 and publish 0.15.0 to maven.

PR #350, opened 2017-05-12
Fix kafka spec failures on travis.

pankajroark:
Kafka spec is pretty flaky on travis. Trying to fix it. At this point the purpose of this PR is to try changes on travis, not ready for review yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ matrix:
   include:
     - scala: 2.11.11
       jdk: oraclejdk8
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage storehaus-kafka/test coverageReport mimaReportBinaryIssues
       after_success:
         - bash <(curl -s https://codecov.io/bash)
     - scala: 2.12.2 
       jdk: oraclejdk8
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,12 +4,12 @@ matrix:
...
     - scala: 2.11.11
       jdk: oraclejdk8
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage storehaus-kafka/test coverageReport mimaReportBinaryIssues
       after_success:
         - bash <(curl -s https://codecov.io/bash)
     - scala: 2.12.2 
       jdk: oraclejdk8
-      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      script: umask 0022 && ./sbt ++$TRAVIS_SCALA_VERSION clean coverage storehaus-kafka/test coverageReport mimaReportBinaryIssues
```
johnynek: so we need to make sure not to merge with this setting.

pankajroark: yes, this is just for trying out faster.

pankajroark:
I think I've found the root cause and fixed it, please take a look. The issue was that on first poll after writing events to Kafka we may not get all. We need to try again a few times if we don't get all events.

pankajroark:
Unfortunately, it looks like even this does not remove the flakiness completely, even though does seem to help.

johnynek:
Thanks for working to get the tests less flaky.

pankajroark:
I think the Kafka tests are much better now. Looking forward to merging this.

johnynek:
:+1:

Next we should use codecov to see where we can easily improve test coverage.

johnynek (approved):

PR #359, opened 2019-03-13
Update finagle version to 7.0.0

nownikhil:
Replaced Ketama client with Memcache client

Bumped finagle version to 7.0.0
`client.release()` method was deprecated and in latest version of finagle release method was removed. 7.0 is the earliest version with close method available.

Also `KetamaClientBuilder` was removed from 7.0 hence replacing it with `Memcached.client`. Destination path for both KetamaClient and Memcached client is same, hence the same `nodeString` works.


ttim (approved):
👍 
wdyt @johnynek ?

nownikhil:
@johnynek Can you have a look at this?

johnynek (approved):
Sorry for the latency. Looks good to me.

PR #360, opened 2019-06-06
Use https instead of http in urls

alanbato:
This PR changes every instance of an http url to its https version for security reasons.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,5 @@
 resolvers ++= Seq(
-  "jgit-repo" at "http://download.eclipse.org/jgit/maven"
+  "jgit-repo" at "https://download.eclipse.org/jgit/maven"
 )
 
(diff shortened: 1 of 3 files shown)


project/plugins.sbt
```diff
@@ -1,5 +1,5 @@
 resolvers ++= Seq(
-  "jgit-repo" at "http://download.eclipse.org/jgit/maven"
```
jenniferyen, line 2: original link broken

jenniferyen (commented):
links resolved except for the one in project/plugins.sbt

sbt
```diff
@@ -80,7 +80,7 @@ make_url () {
   version="$1"
 
   case "$version" in
-        0.7.*) echo "http://simple-build-tool.googlecode.com/files/sbt-launch-0.7.7.jar" ;;
```
jenniferyen, line 83: original link broken

